PR #7, opened 2020-11-30
NIAD-759: Connect GP2GP adaptor to object storage

ElliotB-96:


service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnector.java
```diff
@@ -0,0 +1,57 @@
...
+public class AzureStorageConnector implements StorageConnector {
+
+    private BlobServiceClient blobServiceClient;
+    private BlobContainerClient containerClient;
+
+    private static final String CONTAINER_NAME = "for-nia-testing";
+
+    protected AzureStorageConnector() {
+        String connectionString = System.getenv("AZURE_STORAGE_CONNECTION_STRING");
+        if (connectionString != null && !connectionString.isEmpty()) {
```
mmackiewicz1, line 25: How about using a "StringUtils" class from apache commons?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/LocalMockConnector.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class LocalMockConnector implements StorageConnector {
+
```
mmackiewicz1, line 14: Is this empty line necessary? We already have an indentation.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/S3StorageConnector.java
```diff
@@ -0,0 +1,61 @@
...
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.services.s3.model.ObjectListing;
+import com.amazonaws.services.s3.model.ObjectMetadata;
+import com.amazonaws.services.s3.model.S3Object;
+import com.amazonaws.services.s3.model.S3ObjectSummary;
+
+public class S3StorageConnector implements StorageConnector {
+
+    private final AmazonS3 s3client;
+    private static final String BUCKET_NAME = "for-nia-testing";
```
mmackiewicz1, line 20: Perhaps the bucket name should be in a config file?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorOptions.java
```diff
@@ -0,0 +1,6 @@
+package uk.nhs.adaptors.gp2gp.common.storageconnectors;
+
+public class StorageConnectorOptions {
+    public static final String S3 = "S3";
```
mmackiewicz1, line 4: Could we move these values to a config file?

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Slf4j
+public class StorageConnectorFactoryTest {
+
+    @Test
+    public void checkFactoryReturnsMockConnector() {
+        StorageConnector storageConnector = StorageConnectorFactory.getConfiguredConnector();
+        assertTrue(storageConnector instanceof LocalMockConnector);
```
mmackiewicz1, line 14: We could parametrise this test to assert other connectors.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnectorTest.java
```diff
@@ -0,0 +1,65 @@
...
+    @BeforeAll
+    public static void setupConnector() throws IOException {
+        storageConnector = new AzureStorageConnector();
+        File file = new File(FILE_TO_UPLOAD);
+        InputStream inputStream = FileUtils.openInputStream(file);
+        storageConnector.uploadToStorage(inputStream, SETUP_FILE);
+    }
+
+    @Test
+    public void uploadToAzureStorage() throws IOException {
```
mmackiewicz1, line 35: Could we add some empty lines to make these tests more readable? Also, wha do you think about this test method naming?
"When_UploadingFileToAzureStorage_Expect_FileToBeAvaliableForDownload"

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static uk.nhs.adaptors.gp2gp.common.storageconnectors.StorageConnectorOptions.AZURE_BLOB;
+import static uk.nhs.adaptors.gp2gp.common.storageconnectors.StorageConnectorOptions.S3;
+
+public class StorageConnectorFactory {
+
+    private static StorageConnector storageConnector;
+
+    public static StorageConnector getConfiguredConnector() {
+        String storageMode = System.getenv("STORAGE_SOLUTION");
```
mmackiewicz1, line 11: This environmental variable key could also be stored in config. What do you think?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnectorTest.java
Broom94: I like that naming converntion, junit 5 also has a description for test we could us thoughts on that?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnectorTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class AzureStorageConnectorTest {
+    private static StorageConnector storageConnector;
+    private static final String FILE_TO_UPLOAD = "src/test/resources/test.txt";
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/intTest/resources/downloads/test.txt";
+    private static final String SETUP_FILE = "setupFile.txt";
+    private static final String TEST_UPLOAD = "uploadFile.txt";
```
Broom94, line 24: could you create static methods for these to create objects needed?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/S3StorageConnectorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+public class S3StorageConnectorTest {
+    private static StorageConnector storageConnector;
+    private static final String FILE_TO_UPLOAD = "src/test/resources/test.txt";
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/intTest/resources/downloads/test.txt";
+    private static final String SETUP_FILE = "setupFile.txt";
+    private static final String TEST_UPLOAD = "uploadFile.txt";
```
Broom94, line 26: same again here maybe static method to load them once?

Broom94:
Run ./gradlew check

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnectorTest.java
mattd-kainos: The @mmackiewicz1 naming convention is also what we used on NHAIS

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnectorTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+import org.apache.commons.io.FileUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class AzureStorageConnectorTest {
```
mattd-kainos, line 19: There should be a single integration test that uses the factory or, better yet, dependency injection, to get an instance of StorageConnector. Which backend it tests against should depend entirely upon the configuration used when running the tests. Using the mock/local should be the default, but it should be possible to set environment variables that would cause it to run the AWS or Azure code instead.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactory.java
mattd-kainos: Good use of the factory pattern here! The next step is to use it as a spring factory bean. That will also let us inject the configuration values from application.yml as @mmackiewicz1 suggested.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryTest.java
mattd-kainos: Once StorageConnectorFactory is setup as a spring factory bean we should also be over to override the configuration values - it would be easiest to have a separate StorageConnectorConfiguration class just to hold configuration values for this purpose.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnector.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+
+public interface StorageConnector {
+    void uploadToStorage(InputStream is, String filename) throws IOException;
+    OutputStream downloadFromStorage(String filename) throws IOException;
+    List<String> getFileListFromStorage();
```
mattd-kainos, line 11: We don't have a use case for listing buckets so this should be removed

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnector.java
```diff
@@ -0,0 +1,12 @@
+package uk.nhs.adaptors.gp2gp.common.storageconnectors;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+
+public interface StorageConnector {
+    void uploadToStorage(InputStream is, String filename) throws IOException;
```
mattd-kainos, line 9: We might consider defining a StorageConnectorException to wrap any backend-specific errors that are thrown

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/LocalMockConnectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private static StorageConnector storageConnector;
+    private static final String FILE_TO_UPLOAD = "src/test/resources/test.txt";
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/test/resources/downloads/test.txt";
+    private static final String SETUP_FILE = "setupFile.txt";
+    private static final String TEST_UPLOAD = "uploadFile.txt";
+
+    @BeforeAll
+    public static void setupConnector() throws IOException {
+        storageConnector = StorageConnectorFactory.getConfiguredConnector();
```
mattd-kainos, line 31: create LocalConnector directly

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryTest.java
```diff
@@ -0,0 +1,17 @@
+package uk.nhs.adaptors.gp2gp.common.storageconnectors;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Slf4j
```
mattd-kainos, line 8: Not actually logging anything here - Lombok slf4j annotation not needed

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnector.java
```diff
@@ -0,0 +1,54 @@
...
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
+
+public class AzureStorageConnector implements StorageConnector {
+
+    private BlobServiceClient blobServiceClient;
+    private BlobContainerClient containerClient;
+
+    protected AzureStorageConnector(StorageConnectorConfiguration configuration) {
+        String connectionString = configuration.getAzureConnectionString();
```
mmackiewicz1, line 20: Constructors should be used only to assign fields. Could you define a bean in a configuration class (method with @Bean annotation) which returns a prepared "BlobContainerClient" and then simply inject it here in a constructor?

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryTest.java
```diff
@@ -0,0 +1,34 @@
...
+@SpringBootTest
+public class StorageConnectorFactoryTest {
+
+    @Autowired
+    private StorageConnector storageConnector;
+    @Autowired
+    private StorageConnectorConfiguration configuration;
+
+    @Test
+    public void checkFactoryReturnsCorrectConnector() {
```
mmackiewicz1, line 22: Please, do not write your tests like that. Every single test method should perform one scenario. invoked. Mock "configuration.getPlatform" for "getObject" method for "StorageConnector". Then inject it onto object with [ReflectionTestUtils](https://www.baeldung.com/spring-reflection-test-utils). Then use [Parametrised JUnit tests](https://www.google.pl/search?q=parameterized+test+java&spell=1&sa=X&ved=2ahUKEwijzq2RxK_tAhXWTBUIHStmDpoQBSgAegQICRA1&biw=1440&bih=697) to check if every connector is returned.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @MethodSource("provideConnectorTypesForTest")
+    public void When_PlatformIsLocal_Expect_LocalMockConnector(String platform, Class connectorClass) throws Exception {
+        when(configuration.getPlatform()).thenReturn(platform);
+        when(configuration.getS3AccessKey()).thenReturn("");
+        when(configuration.getS3SecretKey()).thenReturn("");
+
+        StorageConnector storageConnector = storageConnectorFactory.getObject();
+        assertThat(storageConnector).isInstanceOf(connectorClass);
+
+        ReflectionTestUtils.setField(storageConnectorFactory, "storageConnector", null);
```
mmackiewicz1, line 34: This should be sett at the beginning. Does it work without it?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/S3StorageConnector.java
```diff
@@ -0,0 +1,63 @@
...
+import com.amazonaws.services.s3.model.S3Object;
+
+public class S3StorageConnector implements StorageConnector {
+
+    private final AmazonS3 s3client;
+    private final String bucketName;
+
+    protected S3StorageConnector(StorageConnectorConfiguration configuration) {
+        this.bucketName = configuration.getContainerName();
+        AWSCredentials credentials = new BasicAWSCredentials(
```
mattd-kainos, line 24: For AWS leave out credentials - the AmazonS3 client will read them from the local environments. This lets us just give our Fargate task an IaM role with access to the S3 bucket without managing the keys ourselves.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/S3StorageConnector.java
```diff
@@ -0,0 +1,63 @@
...
+            configuration.getS3AccessKey(),
+            configuration.getS3SecretKey()
+        );
+        this.s3client = AmazonS3ClientBuilder
+            .standard()
+            .withCredentials(new AWSStaticCredentialsProvider(credentials))
+            .withRegion(Regions.EU_WEST_2)
```
mattd-kainos, line 31: Same with region, we won't always be in EU_WEST_2 - omit and it let it be ready from the local environment. You may need to define these locally or setup the aws cli for it to work from your local machine though.

service/src/main/resources/application.yml
```diff
@@ -8,4 +8,9 @@ management:
       path-mapping.health: healthcheck
       exposure:
         include: info, health, metrics, mappings
-
+storage:
+  platform: ${STORAGE_SOLUTION:LocalMock}
+  containerName: ${CONTAINER_NAME:for-nia-testing}
+  s3AccessKey: ${AWS_ACCESS_KEY_ID:for-nia-testing}
```
mattd-kainos, line 14: remove the S3 values, you won't need them

service/src/main/resources/application.yml
```diff
@@ -8,4 +8,9 @@ management:
       path-mapping.health: healthcheck
       exposure:
         include: info, health, metrics, mappings
-
+storage:
+  platform: ${STORAGE_SOLUTION:LocalMock}
```
mattd-kainos, line 12: storage.platform, $STORAGE_SOLUTION

names are inconsistent.

storage.type and $GP2GP_STORAGE_TYPE?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.beans.factory.FactoryBean;
+
+@Setter
+public class StorageConnectorFactory implements FactoryBean<StorageConnector> {
+
+    private static StorageConnector storageConnector;
+    private StorageConnectorConfiguration configuration;
+
+    @Override
+    public StorageConnector getObject() throws Exception {
```
Broom94, line 16: Do we need this exception?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactory.java
```diff
@@ -0,0 +1,41 @@
...
+                    storageConnector = new S3StorageConnector(configuration);
+                    break;
+                case AZURE:
+                    storageConnector = new AzureStorageConnector(configuration);
+                    break;
+                default:
+                    storageConnector = new LocalMockConnector();
+            }
+        }
+        return storageConnector;
```
Broom94, line 29: This could pass NPE

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactory.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public Class<?> getObjectType() {
+        return StorageConnector.class;
+    }
+
+    @Override
+    public boolean isSingleton() {
+        return true;
```
Broom94, line 39: No need for this method, default is set to true in proxyFactoryBean

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Bean(name = "storage-connector")
+    public StorageConnectorFactory storageConnectorFactory() {
+        StorageConnectorFactory factory = new StorageConnectorFactory();
+        factory.setConfiguration(configuration);
+        return factory;
+    }
+
+    @Bean
+    public StorageConnector storageConnector() throws Exception {
```
Broom94, line 21: Marcin&Thomas: This will not be needed once above exception is removed

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactoryTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class StorageConnectorFactoryTest {
+
+    @InjectMocks
+    private StorageConnectorFactory storageConnectorFactory;
+    @Mock
+    private StorageConnectorConfiguration configuration;
+
+    @ParameterizedTest
+    @MethodSource("provideConnectorTypesForTest")
+    public void When_PlatformIsLocal_Expect_LocalMockConnector(String platform, Class connectorClass) throws Exception {
```
mmackiewicz1, line 26: After consulting this with @Broom94 we decided that unit test would not correctly test the logic behind the factory operation. A component test would be more fitting in this case.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+public class StorageConnectorTest {
+
+    @Autowired
+    private StorageConnector storageConnector;
+
+    private static final String FILE_TO_UPLOAD = "src/intTest/resources/test.txt";
```
Broom94, line 24: @mmackiewicz1 & @Broom94 use getClass().getResource(test.txt).getPath()

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+        FileUtils.copyToFile(inputStream, expectedDownload);
+        assertEquals(FileUtils.readLines(upload), FileUtils.readLines(expectedDownload));
+    }
+
+    @AfterAll
+    public static void deleteDownloadedFile() {
+        new File(FILE_EXPECTED_DOWNLOAD).delete();
+    }
+
+    private static File getFileToUpload() {
```
Broom94, line 53: @mmackiewicz1 & @Broom94 please rename as this isn't a getter class, more of initialiser/constructor.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+public class StorageConnectorTest {
+
+    @Autowired
+    private StorageConnector storageConnector;
+
+    private static final String FILE_TO_UPLOAD = "src/intTest/resources/test.txt";
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/intTest/resources/downloads/test.txt";
```
Broom94, line 25: @mmackiewicz1 change file name to reflect contents

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String SETUP_FILE = "setupFile.txt";
+
+    @BeforeEach
+    public void setupConnector() throws IOException {
+        File file = getFileToUpload();
```
Broom94, line 30: private static final File file = load<FileName>FromFilePath

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/LocalMockConnectorTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class LocalMockConnectorTest {
+
+    private static StorageConnector storageConnector;
+    private static final String FILE_TO_UPLOAD = "src/test/resources/test.txt";
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/test/resources/downloads/test.txt";
+    private static final String SETUP_FILE = "setupFile.txt";
```
Broom94, line 21: @mmackiewicz1 & @Broom94 create const class for strings reused

service/src/test/resources/application.yml
```diff
@@ -0,0 +1,16 @@
...
+      base-path: /
+      path-mapping.health: healthcheck
+      exposure:
+        include: info, health, metrics, mappings
+storage:
+  platform: ${STORAGE_SOLUTION:LocalMock}
+  containerName: ${CONTAINER_NAME:for-nia-testing}
+  s3AccessKey: ${AWS_ACCESS_KEY_ID:for-nia-testing}
+  s3SecretKey: ${AWS_SECRET_ACCESS_KEY:for-nia-testing}
+  azureConnectionString: ${AZURE_STORAGE_CONNECTION_STRING:for-nia-testing}
```
Broom94, line 16: Add new variable to readme

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
ElliotB-96: Discussed with Thomas, leaving for further stories

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/LocalMockConnectorTest.java
ElliotB-96: Strings no longer being reused mock test class removed

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorFactory.java
ElliotB-96: Reviewed with Thomas leaving as is

README.md
```diff
@@ -13,17 +13,23 @@ The "National Integration Adaptor - GP2GP" implements a GP2GP 2.2 producer using
...
 
-| Environment Variable               | Default                   | Description 
+| Environment Variable               | Default                   | Description
 | -----------------------------------|---------------------------|-------------
 | GP2GP_SERVER_PORT                  | 8080                      | The port on which the SCR API will run
 | GP2GP_LOGGING_LEVEL                | INFO                      | Application logging level. One of: DEBUG, INFO, WARN, ERROR. The level DEBUG **MUST NOT** be used when handling live patient data.
 | GP2GP_LOGGING_FORMAT               | (*)                       | Defines how to format log events on stdout
+| GP2GP_STORAGE_TYPE                 | LocalMock                 | Defines the storage solution being used (S3, Azure, LocalMock)
+| GP2GP_STORAGE_CONTAINER_NAME       | for-nia-testing           | Defines the name of the BlobStorage container on Azure or bucket on S3
+| AZURE_STORAGE_CONNECTION_STRING    |                           | Defines the connection string used to connect to azure blob storage
```
mattd-kainos, line 29: Prefix with GP2GP since it is "our" variable. `GP2GP_AZURE_STORAGE_CONNECTION_STRING` they get a bit long winded but 🤷

service/build.gradle
```diff
@@ -30,6 +30,10 @@ dependencies {
 	implementation 'ch.qos.logback.contrib:logback-json-classic:0.1.5'
 	implementation 'ch.qos.logback.contrib:logback-jackson:0.1.5'
 	implementation 'org.springframework.boot:spring-boot-starter-web'
+	implementation 'com.amazonaws:aws-java-sdk-s3:1.11.907'
+	implementation 'com.azure:azure-storage-blob:12.9.0'
+	implementation 'commons-io:commons-io:2.8.0'
+	implementation 'org.apache.commons:commons-lang3:3.0'
```
mattd-kainos, line 36: 3.0 is not the latest version 
http://commons.apache.org/proper/commons-lang/

service/src/test/resources/application.yml
```diff
@@ -0,0 +1,14 @@
...
+  port: ${GP2GP_SERVER_PORT:8080}
+
+management:
+  endpoints:
+    web:
+      base-path: /
+      path-mapping.health: healthcheck
+      exposure:
+        include: info, health, metrics, mappings
+storage:
```
mattd-kainos, line 11: This these should be gp2gp.storage? Would avoid any potential naming conflicts if there is some other spring module out there using the "storage" root config property

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnector.java
```diff
@@ -0,0 +1,39 @@
...
+    @Autowired
+    private BlobContainerClient containerClient;
+
+    protected AzureStorageConnector() {}
+
+    @Override
+    public void uploadToStorage(InputStream is, String filename) throws StorageConnectorException {
+        try {
+            BlobClient blobClient = containerClient.getBlobClient(filename);
+            blobClient.upload(is, is.available());
```
mattd-kainos, line 22: Have a look at the javadoc for InputStream.available()

> Returns an estimate of the number of bytes that can be read (or skipped over) from this input stream without blocking

The impact of using this method would mean we might not upload the entire data stream!

I think our interface is going to need an additional 'length' parameter. In the future when we're streaming http response bodies directly into storage we can use the Content-Length header to set this parameter

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnector.java
```diff
@@ -0,0 +1,39 @@
...
+            throw new StorageConnectorException("Error occurred uploading to Azure Storage", ioException);
+        }
+    }
+
+    @Override
+    public OutputStream downloadFromStorage(String filename) throws StorageConnectorException {
+        try {
+            BlobClient blobClient = containerClient.getBlobClient(filename);
+            OutputStream outputStream = new ByteArrayOutputStream();
+            blobClient.openInputStream().transferTo(outputStream);
```
mattd-kainos, line 33: Creating a ByteArrayOutputStream and transferring blobClient.openInputStream() into it means we're loading the entire file into memory. What happens if the file is 1GB? What happens if the adaptor is processing many of these 1GB files at the same time?

From the story test and tech notes:

> The adaptor will not store any files locally and should avoid loading entire files into memory

Just return the OutputStream from blobClient.openInputStream() ;)

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/BlobContainerClientConfig.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
+import static uk.nhs.adaptors.gp2gp.common.storageconnectors.StorageConnectorOptions.AZURE;
+
+@Configuration
+@ComponentScan(basePackageClasses = StorageConnectorFactory.class)
```
mattd-kainos, line 14: What is the ComponentScan doing here. Is it needed? Spring Boot scan's the application's package and sub-packages already.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/S3StorageConnector.java
```diff
@@ -0,0 +1,52 @@
...
+        this.s3client = AmazonS3ClientBuilder
+            .standard()
+            .build();
+    }
+
+    @Override
+    public void uploadToStorage(InputStream is, String filename) throws StorageConnectorException {
+        try {
+            ObjectMetadata metadata = new ObjectMetadata();
+            metadata.setContentLength(is.available());
```
mattd-kainos, line 28: Same as above InputStream.available() may not give the entire length of the content. Use a length parameter on uploadToStorage method.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/S3StorageConnector.java
```diff
@@ -0,0 +1,52 @@
...
+        } catch (IOException ioException) {
+            throw new StorageConnectorException("Error occurred uploading to S3 Bucket", ioException);
+        }
+    }
+
+    @Override
+    public OutputStream downloadFromStorage(String filename) throws StorageConnectorException {
+        try {
+            S3Object s3Object = s3client.getObject(bucketName, filename);
+            InputStream in = s3Object.getObjectContent();
```
mattd-kainos, line 44: Just return s3Object.getObjectContent() for reasons mentioned in Azure section.

Will need to change interface signature to return InputStream. I'm probably to blame for it being OutputStream!

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorOptions.java
mattd-kainos: I think they are ok enumerated in code. You could consider changing this to a Java enum though.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/AzureStorageConnector.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import org.springframework.beans.factory.annotation.Autowired;
+import com.azure.storage.blob.BlobClient;
+import com.azure.storage.blob.BlobContainerClient;
+
+public class AzureStorageConnector implements StorageConnector {
+
+    @Autowired
+    private BlobContainerClient containerClient;
```
mattd-kainos, line 14: Are you sure this works? This property is autowired but the StorageConnectorFactor instantiates it with `new`. I don't see how this field would be injected.

ElliotB-96: I have just tested this and it still appears to be working correctly.

mattd-kainos: Nevermind, this is some spring magic. Dependency injection happens after FactoryBean returns its instance apparently

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+public class StorageConnectorTest {
+
+    @Autowired
+    private StorageConnector storageConnector;
+
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/intTest/resources/downloads/test.txt";
```
mattd-kainos, line 24: Should use a classpath reference rather than a relative file path

service/src/intTest/resources/application.yml
```diff
@@ -16,4 +19,11 @@ mongodb:
...
-  cosmosDbEnabled: ${GP2GP_COSMOS_DB_ENABLED:false}
\ No newline at end of file
+  cosmosDbEnabled: ${GP2GP_COSMOS_DB_ENABLED:false}
+        include: info, health, metrics, mappings
+
+gp2gp:
+  storage:
+    type: ${GP2GP_STORAGE_TYPE:LocalMock}
+    containerName: ${GP2GP_STORAGE_CONTAINER_NAME:for-nia-testing}
+    azureConnectionString: ${GP2GP_AZURE_STORAGE_CONNECTION_STRING:for-nia-testing}
```
mattd-kainos, line 29: Maybe leave off the "for-nia-testing" defaults here? These just happen to be the name we picked for the storage / S3 buckets but are not needed for "LocalMock".

service/src/main/resources/application.yml
```diff
@@ -8,7 +8,11 @@ management:
       path-mapping.health: healthcheck
       exposure:
         include: info, health, metrics, mappings
-
+gp2gp:
+  storage:
+    type: ${GP2GP_STORAGE_TYPE:LocalMock}
+    containerName: ${GP2GP_STORAGE_CONTAINER_NAME:for-nia-testing}
+    azureConnectionString: ${GP2GP_AZURE_STORAGE_CONNECTION_STRING:for-nia-testing}
```
mattd-kainos, line 15: as above, remove "for-nia-testing" defaults

service/src/test/resources/application.yml
```diff
@@ -8,4 +8,20 @@ management:
       path-mapping.health: healthcheck
       exposure:
         include: info, health, metrics, mappings
+gp2gp:
+  storage:
+    type: ${GP2GP_STORAGE_TYPE:LocalMock}
+    containerName: ${GP2GP_STORAGE_CONTAINER_NAME:for-nia-testing}
+    azureConnectionString: ${GP2GP_AZURE_STORAGE_CONNECTION_STRING:for-nia-testing}
```
mattd-kainos, line 15: and here

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
mmackiewicz1: This is an expected url for a downloaded file which is not yet present. Therefore classpath reference will cause a null pointer when attempting to construct the URI. We could treat is as an assert value.

service/src/intTest/resources/application.yml
mmackiewicz1: Ok

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String SETUP_FILE_TXT = "setupFile.txt";
+    private File uploadedFile;
+    private File downloadedFile;
+
+    @BeforeEach
+    public void setupConnector() throws IOException {
+        String fileToUpload = StorageConnectorTest.class.getResource("/test.txt").getPath();
+        uploadedFile = new File(fileToUpload);
```
michael-kainos, line 37: This instance method writes to a static field. This is tricky to get correct if multiple instances are being manipulated, and generally bad practice.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void When_DownloadingFileFromStorage_Expect_FileContentToMatch() throws IOException {
+        InputStream inputStream = new ByteArrayInputStream(storageConnector.downloadFromStorage(SETUP_FILE_TXT).readAllBytes());
+        downloadedFile = new File(FILE_EXPECTED_DOWNLOAD);
+        FileUtils.copyToFile(inputStream, downloadedFile);
+
+        assertThat(FileUtils.contentEquals(uploadedFile, downloadedFile), is(true));
+    }
+
+    @AfterEach
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED_BAD_PRACTICE")
```
michael-kainos, line 52: Why are we ignoring bad practice here? This method returns a value that is not checked.

michael-kainos (changes requested):

service/src/intTest/java/uk/nhs/adaptors/gp2gp/common/storageconnectors/StorageConnectorTest.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String FILE_EXPECTED_DOWNLOAD = "src/intTest/resources/downloads/test.txt";
+    private static final String SETUP_FILE_TXT = "setupFile.txt";
+    private File uploadedFile;
+    private File downloadedFile;
+
+    @BeforeEach
+    public void setupConnector() throws IOException {
+        String fileToUpload = StorageConnectorTest.class.getResource("/test.txt").getPath();
+        uploadedFile = new File(fileToUpload);
+        InputStream inputStream = FileUtils.openInputStream(uploadedFile);
```
michael-kainos, line 38: Where are we closing this InputStream object?

michael-kainos (approved):

mattd-kainos (approved):

michael-kainos (approved):

PR #24, opened 2020-12-21
NIAD-796 Added wiremock with support for docker run

mmackiewicz1:


bartek-sarul (approved):

mattd-kainos (approved):
Added a README entry. The stub interactions is not GPC or SDS, but our story doesn't require that

bartek-sarul (approved):

PR #32, opened 2021-01-08
NIAD-809 Adding logging level and add common identifiers to logging format

z-uddin:


ElliotB-96 (dismissed):
Looks good to me, can also see both conversation id and task id in the logs.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/ConversationIdService.java
```diff
@@ -0,0 +1,21 @@
+package uk.nhs.adaptors.gp2gp.common.service;
+
+import org.slf4j.MDC;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConversationIdService {
```
mattd-kainos, line 7: I'd consider making a single MDC service with `applyConversationId` and `applyTaskId` methods and a single `resetAll` method. That encapsulates all of the code related to logging diagnostic identifiers in a single place.

I'd question whether we need getters at all. They were used in NHAIS but I'm not sure that was a great design choice.

mattd-kainos (changes requested):
Looks good, all of the ids are set and reset from the correct places. Recommend having a single service to manage all mapped diagnostic values.

ElliotB-96 (approved):

mattd-kainos (approved):
Looks good! Also did a quick test and they values are included in the logs as expected 

```
"mdc":{"ConversationId":"1f712f0c-8d56-40ab-96bc-5eb6a31ab532","TaskId":"696b4617-59c4-44f6-aded-1c40ec012b5b"}
```

PR #36, opened 2021-01-13
Feature/niad 790 mock gpc api wiremock fix

mmackiewicz1:


Broom94 (dismissed):

docker/wiremock/stubs/mappings/retreiveNoneExistingPatientDocument.json
```diff
@@ -14,7 +14,6 @@
       "Content-Length":"291",
```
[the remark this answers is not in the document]
mmackiewicz1: I just checked the response length for 404 response from https://orange.testlab.nhs.uk/ and length there is even shorter - "280". I am not sure if this is the case.

ElliotB-96 (dismissed):

docker/wiremock/stubs/mappings/retreiveNoneExistingPatientDocument.json
```diff
@@ -11,10 +11,9 @@
       "Server":"nginx",
       "Date":"Tue, 12 Jan 2021 11:36:57 GMT",
       "Content-Type":"application/fhir+json;charset=UTF-8",
-      "Content-Length":"291",
+      "Content-Length":"528",
```
mattd-kainos, line 14: You normally would not hardcode a content-length unless you're trying to test an invalid response header. Surely wiremock calculates this by default?

mattd-kainos (changes requested):
Does wiremock calculate a content-length for you if you omit it?

docker/wiremock/stubs/mappings/retreiveNoneExistingPatientDocument.json
mmackiewicz1: I think it will be the best to also remove the content length. It is not necessary for a response body to be returned.

mattd-kainos (approved):

Broom94 (approved):

PR #39, opened 2021-01-14
NIAD-788: GPC get access structured

Broom94:


service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtils.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class StorageUtils {
+    public int getInputStreamSize(InputStream inputStream) throws IOException {
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        int b;
+        while ((b = inputStream.read()) != -1) {
+            os.write(b);
```
Broom94, line 15: Don't think this is 100%, could we just you string.length?

service/src/main/resources/application.yml
```diff
@@ -32,3 +33,8 @@ gp2gp:
     autoIndexCreation: ${GP2GP_MONGO_AUTO_INDEX_CREATION:true}
     ttl: ${GP2GP_MONGO_TTL:P7D}
     cosmosDbEnabled: ${GP2GP_COSMOS_DB_ENABLED:false}
+
+gpConnect:
+  url: "https://orange.testlab.nhs.uk/B82617/STU3/1/gpconnect/fhir"
+  endpoint: "/Patient/$gpc.getstructuredrecord"
+  host: "orange.testlab.nhs.uk"
```
Broom94, line 40: add temp env variable

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,37 @@
...
+    public void handleStructuredRecord(GpcStructuredResponseObject response) throws IOException {
+        String jsonStringResponse = objectMapper.writeValueAsString(response);
+        var responseInputStream = new ByteArrayInputStream(jsonStringResponse.getBytes(UTF_8));
+
+        var storageConnector = Optional.ofNullable(storageConnectorFactory.getObject())
+            .orElseThrow(() ->
+                new TaskHandlerException("No storage connector available"));
+
+        var length = storageUtils.getInputStreamSize(responseInputStream);
+        storageConnector.uploadToStorage(responseInputStream, length, response.getConversationId() + "_gpc_structured.json");
```
mmackiewicz1, line 35: "_gpc_structured.json" could be moved to a static constant.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtils.java
```diff
@@ -0,0 +1,19 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class StorageUtils {
+    public int getInputStreamSize(InputStream inputStream) throws IOException {
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        int b;
```
mmackiewicz1, line 13: Please, set some more descriptive names to variables.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/request/EhrExtractRequestHandler.java
```diff
@@ -103,4 +109,23 @@ private String getRequiredValue(Document xml, String xpath) {
...
+        if (ehrStatus.isPresent()) {
+            return ehrStatus.get();
+        } else {
+            throw new RuntimeException("No EHR Extract Status for ConversationId: " + structuredTaskDefinition.getConversationId());
+        }
+    }
+
+    public void updateEhrExtractStatusAccessStructured(GetGpcStructuredTaskDefinition structuredTaskDefinition,
+        EhrExtractStatus ehrExtractStatus) {
+        var accessStructured = new EhrExtractStatus.GpcAccessStructured(ehrExtractStatus.getConversationId() + "_gpc_structured.json", now(),
```
mmackiewicz1, line 125: "_gpc_structured.json" could move to a static constant.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcClient.java
```diff
@@ -0,0 +1,37 @@
...
+    private final GpcConfiguration gpcConfiguration;
+
+    public GpcStructuredResponseObject getStructuredRecord(WebClient.RequestHeadersSpec<? extends WebClient.RequestHeadersSpec<?>> request,
+        GetGpcStructuredTaskDefinition structuredTaskDefinition, EhrExtractStatus ehrExtractStatus) {
+        LOGGER.info("Gpc Structured Request, toASID: {}, fromASID: {}, Gpc Endpoint: {}",
+            ehrExtractStatus.getEhrRequest().getToAsid(),
+            ehrExtractStatus.getEhrRequest().getFromAsid(),
+            gpcConfiguration.getUrl() + gpcConfiguration.getEndpoint());
+
+        var responseString =  request
```
mmackiewicz1, line 24: Unnecessary space.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcRequestBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final int NUMBER_OF_RECENT_CONSULTANTS = 3;
+
+    private final IParser fhirParser;
+    private final GpcTokenBuilder gpcTokenBuilder;
+    private final GpcConfiguration gpcConfiguration;
+    private final GpcWebClientFilter gpcWebClientFilter;
+
+    public Parameters buildGetStructuredRecordRequestBody(GetGpcStructuredTaskDefinition structuredTaskDefinition) {
+        return new Parameters()
+            .addParameter(new Parameters.ParametersParameterComponent()
```
mmackiewicz1, line 60: I am wondering if we could create a method which would take the name and the value as parameters to reduce the size of this initialisation.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcRequestBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+            .method(HttpMethod.POST)
+            .uri(gpcConfiguration.getEndpoint());
+
+        var requestBody = fhirParser.encodeResourceToString(requestBodyParameters);
+        BodyInserter<Object, ReactiveHttpOutputMessage> bodyInserter
+            = BodyInserters.fromValue(requestBody);
+
+        var request = ehrExtractStatus.getEhrRequest();
+
+        return uri
```
mmackiewicz1, line 103: Could we move this part to a separate method?

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcRequestBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+        return SslContextBuilder
+            .forClient()
+            .trustManager(InsecureTrustManagerFactory.INSTANCE)
+            .build();
+    }
+
+    private org.springframework.web.reactive.function.client.WebClient buildWebClient(HttpClient httpClient) {
+        return org.springframework.web.reactive.function.client.WebClient
+            .builder()
+            .exchangeStrategies(
```
mmackiewicz1, line 126: That is a lot of indentations. Could we move them to a separate method?

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcRequestBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private SslContext buildSSLContext() throws SSLException {
+        return SslContextBuilder
+            .forClient()
+            .trustManager(InsecureTrustManagerFactory.INSTANCE)
+            .build();
+    }
+
+    private org.springframework.web.reactive.function.client.WebClient buildWebClient(HttpClient httpClient) {
```
mmackiewicz1, line 123: Is there a problem with an import? Does a WebClient class really need to be used with each package?

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcService.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public void handleStructureTask(GetGpcStructuredTaskDefinition structuredTaskDefinition) throws IOException {
+        var ehrExtractStatus = ehrExtractRequestHandler.getEhrExtractStatus(structuredTaskDefinition);
+
+        var requestBodyParameters = gpcRequestBuilder.buildGetStructuredRecordRequestBody(structuredTaskDefinition);
+        var request = gpcRequestBuilder.buildGetStructuredRecordRequest(requestBodyParameters, ehrExtractStatus);
+        var response = gpcClient.getStructuredRecord(request, structuredTaskDefinition, ehrExtractStatus);
+
+        storageConnectorService.handleStructuredRecord(response);
+        ehrExtractRequestHandler.updateEhrExtractStatusAccessStructured(structuredTaskDefinition, ehrExtractStatus);
```
mmackiewicz1, line 33: 👍

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcTemplateUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public static String fillTemplate(Mustache template, Object content) {
+        StringWriter writer = new StringWriter();
+        String data = "";
+
+        try {
+            template.execute(writer, content).flush();
+            data += writer.toString();
+        } catch (IOException e) {
+            e.printStackTrace();
```
mmackiewicz1, line 28: It is not a good practice to print a stack trace. Try making an error log, please.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcTokenBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+public class GpcTokenBuilder {
+    private static final Mustache JWT_HEADER_TEMPLATE = GpcTemplateUtils.loadTemplate("jwt.header.mustache");
+    private static final Mustache JWT_PAYLOAD_TEMPLATE = GpcTemplateUtils.loadTemplate("jwt.payload.mustache");
+    private static final int EXPIRY_TIME_ADDITION = 300000;
+    private static final int MILLISECOND_DIVISION = 1000;
+
+    private final GpcConfiguration gpcConfiguration;
+
+    public String buildToken(String odsFromCode) {
+
```
mmackiewicz1, line 31: Please remove the empty line.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtilsTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.junit.jupiter.api.Test;
+
+public class StorageUtilsTest {
+
+    @Test
+    public void When_FileInputSizeIsNine_Then_ReturnNine() throws IOException {
+        StorageUtils storageUtils = new StorageUtils();
+        String test = "123456789";
+        InputStream is = new ByteArrayInputStream(test.getBytes());
```
mmackiewicz1, line 17: "is" sounds quite generic. How about "inputStream"?

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcWebClientFilter.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
+
+import lombok.extern.slf4j.Slf4j;
+import reactor.core.publisher.Mono;
+
+@Slf4j
+@Component
+public class GpcWebClientFilter {
+    public ExchangeFilterFunction errorHandlingFilter() {
+        return ExchangeFilterFunction.ofResponseProcessor(clientResponse -> {
```
mmackiewicz1, line 13: This curly bracket content could be moved to a separate method.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtilsTest.java
Broom94: Is this needed? I think we can just use string.length, what do you think

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcWebClientFilter.java
```diff
@@ -0,0 +1,24 @@
...
+            clientResponse.statusCode();
+            if (clientResponse.statusCode().is5xxServerError() || clientResponse.statusCode().is4xxClientError()) {
```
mmackiewicz1, line 15: "clientResponse.statusCode();" and "if" are in the same indentation. There should be an empty line here.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtils.java
Broom94: Is this util needed? I think we can just use string.length, what do you think

mmackiewicz1: if it is exactly the same size then why not?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtils.java
mmackiewicz1: If it is the same size then why not?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.nhs.adaptors.gp2gp.common.task.TaskHandlerException;
+import uk.nhs.adaptors.gp2gp.gpc.GpcConfiguration;
+import uk.nhs.adaptors.gp2gp.gpc.GpcStructuredResponseObject;
+
+@Service
+@AllArgsConstructor
+public class StorageConnectorService {
+    private final StorageConnectorFactory storageConnectorFactory;
+    private final ObjectMapper objectMapper;
+    private final GpcConfiguration gpcConfiguration;
```
mattd-kainos, line 23: gpcConfiguration is unused

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcService.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import uk.nhs.adaptors.gp2gp.common.storage.StorageConnectorService;
+import uk.nhs.adaptors.gp2gp.ehr.request.EhrExtractRequestHandler;
+
+@Service
+@Slf4j
+@RestController
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class GpcService {
```
mattd-kainos, line 18: Could this class be inlined into the task executor? All the task executor does currently is call this class' sole method. Might be an unnecessary level in indirection.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcService.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private final GpcClient gpcClient;
+    private final GpcRequestBuilder gpcRequestBuilder;
+    private final StorageConnectorService storageConnectorService;
+    private final EhrExtractRequestHandler ehrExtractRequestHandler;
+
+    public void handleStructureTask(GetGpcStructuredTaskDefinition structuredTaskDefinition) throws IOException {
+        var ehrExtractStatus = ehrExtractRequestHandler.getEhrExtractStatus(structuredTaskDefinition);
```
mattd-kainos, line 26: Instead of fetching the ehrExtractStatus could we include toAsid and fromAsid in task definition? Avoid making a database request that is not needed

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,37 @@
...
+    private final StorageUtils storageUtils;
+
+    public void handleStructuredRecord(GpcStructuredResponseObject response) throws IOException {
+        String jsonStringResponse = objectMapper.writeValueAsString(response);
+        var responseInputStream = new ByteArrayInputStream(jsonStringResponse.getBytes(UTF_8));
+
+        var storageConnector = Optional.ofNullable(storageConnectorFactory.getObject())
```
mattd-kainos, line 30: This null check is not needed. StorageConnectorFactory::getObject cannot return null.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,37 @@
...
+            .orElseThrow(() ->
+                new TaskHandlerException("No storage connector available"));
+
+        var length = storageUtils.getInputStreamSize(responseInputStream);
```
mattd-kainos, line 34: You don't need to copy the stream to get the length. I few lines above you've taken the String to bytes: `jsonStringResponse.getBytes(UTF_8)`. The length of this byte array is the length of your stream.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
Broom94: That's what I thought but spotbugs has other ideas

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageUtils.java
mattd-kainos: See my comment in the class that uses this. We don't need this StorageUtils at all.

mattd-kainos (changes requested):
changes requested / still reviewing

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/request/EhrExtractRequestHandler.java
```diff
@@ -103,4 +109,24 @@ private String getRequiredValue(Document xml, String xpath) {
         }
         return value;
     }
+
+    public EhrExtractStatus getEhrExtractStatus(GetGpcStructuredTaskDefinition structuredTaskDefinition) {
```
mattd-kainos, line 113: These methods don't belong in EhrExtractRequestHandler. The purpose of this class is to handle the "EHR Extract Request" - a specific message we receive from the requesting practice via Spine. It is not related to to fetching the patient's data from GP connect

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/request/EhrExtractRequestHandler.java
```diff
@@ -103,4 +109,24 @@ private String getRequiredValue(Document xml, String xpath) {
...
+    }
+
+    public void updateEhrExtractStatusAccessStructured(GetGpcStructuredTaskDefinition structuredTaskDefinition,
+        EhrExtractStatus ehrExtractStatus) {
+        var accessStructured = new EhrExtractStatus.GpcAccessStructured(ehrExtractStatus.getConversationId()
+            + "_gpc_structured.json", now(),
+            structuredTaskDefinition.getTaskId());
+        ehrExtractStatus.setGpcAccessStructured(accessStructured);
+        ehrExtractStatus.setUpdatedAt(now());
+        ehrExtractStatusRepository.save(ehrExtractStatus);
```
mattd-kainos, line 130: This needs to be an update statement not a get/replace. What happens if two instance are updating this document at the same time? We don't have optimistic locking. The other party's updates would be lost!

Broom94: Discussed on call with @mattd-kainos  @mmackiewicz1 pending test

README.md
```diff
@@ -36,6 +36,9 @@ Variables without a default value and not marked optional, *MUST* be defined for
 | GP2GP_AMQP_MAX_REDELIVERIES          | 3                         | The number of times an message will be retried to be delivered to consumer. After exhausting all retires, it will be put on DLQ.<queue_name> dead letter queue
 | GP2GP_MHS_INBOUND_QUEUE              | inbound                   | Name of the queue for MHS inbound
 | GP2GP_TASK_QUEUE                     | gp2gpTaskQueue            | Defines name of internal taskQueue.
+| GP2GP_GPC_GET_URL                    | 8080                      | The URL used for GP Connect requests.
+| GP2GP_GPC_GET_STRUCTURED_ENDPOINT    | 8080                      | The endpoiint for GP Connect Get Structured Access. 
+| GP2GP_GPC_HOST                       | 8080                      | The host used for GP Connect requests.
```
mattd-kainos, line 41: The Default for these is not 8080. I'm fine with leaving them blank in the README.

e2e-tests/src/test/java/uk/nhs/adaptors/gp2gp/e2e/AwaitHelper.java
```diff
@@ -9,9 +9,9 @@
 
 public class AwaitHelper {
 
-    protected static final int WAIT_FOR_IN_SECONDS = 10;
+    protected static final int WAIT_FOR_IN_SECONDS = 30;
     protected static final int POLL_INTERVAL_MS = 100;
-    protected static final int POLL_DELAY_MS = 10;
+    protected static final int POLL_DELAY_MS = 5000;
```
mattd-kainos, line 14: This change will make every await call take at least 5 seconds. Our tests will be very slow.

I know why you've done this though. The database is not updated to its "final" state right away.

Leave these settings as they were and we will find a better way to write the test.

e2e-tests/src/test/java/uk/nhs/adaptors/gp2gp/e2e/EhrExtractTest.java
```diff
@@ -45,6 +49,10 @@ public void When_extractRequestReceived_Expect_ExtractStatusAddedToDatabase() th
         assertThat(ehrRequest.get("toAsid")).isEqualTo(TO_ASID);
         assertThat(ehrRequest.get("fromOdsCode")).isEqualTo(FROM_ODS_CODE);
         assertThat(ehrRequest.get("toOdsCode")).isEqualTo(TO_ODS_CODE);
+        var accessStructured = (Document) ehrExtractStatus.get(GPC_ACCESS_STRUCTURED);
```
mattd-kainos, line 52: What if we split our test/assertion into two helper methods:
assertThatInitialRecordWasCreated
assertThatAccessStructuredWasFetched

within `assertThatAccessStructuredWasFetched` we can have a second `waitFor(() -> ` that waits until the ehrRequest exists.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.AllArgsConstructor;
+import uk.nhs.adaptors.gp2gp.gpc.GpcStructuredResponseObject;
+
+@Service
+@AllArgsConstructor
+public class StorageConnectorService {
+    private final StorageConnectorFactory storageConnectorFactory;
+    private final ObjectMapper objectMapper;
+
+    @SuppressFBWarnings(
```
mattd-kainos, line 24: Doh! I totally missed the actual issue here.

Don't ever use `StorageConnectorFactory` directly. This is something Spring uses behind-the-scenes to create an instance of StorageConnector.

Instead just declare `private final StorageConnector storageConnector;` and let spring inject the dependency.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,36 @@
...
+        value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
+        justification = "SpotBugs issue with fix not yet released https://github.com/spotbugs/spotbugs/issues/456")
+    public void handleStructuredRecord(GpcStructuredResponseObject response) throws IOException {
+        String jsonStringResponse = objectMapper.writeValueAsString(response);
+        var responseInputStream = new ByteArrayInputStream(jsonStringResponse.getBytes(UTF_8));
+
+        var storageConnector = storageConnectorFactory.getObject();
+
+        var length = jsonStringResponse.getBytes(UTF_8).length;
```
mattd-kainos, line 33: Don't call getBytes(UTF_8) twice to just get the length. For a large payload this is a potentially expensive operation!

Why not declare a `var responseBytes = jsonStringResponse.getBytes(UTF_8)` so you can reuse it?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/task/TaskExecutor.java
```diff
@@ -1,8 +1,10 @@
 package uk.nhs.adaptors.gp2gp.common.task;
 
+import java.io.IOException;
+
 public interface TaskExecutor<T extends TaskDefinition> {
 
     Class<T> getTaskType();
 
-    void execute(T taskDefinition);
+    void execute(T taskDefinition) throws IOException;
```
mattd-kainos, line 9: Don't add the checked exception here. IOException is also a very broad type - who know why it might be throwing an IOException here?

I have not gotten far enough to see why this is needed, but if this is something we can intercept and at least explain we should do a catch-and-release with our error message and RuntimeException subclass.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,36 @@
...
+@AllArgsConstructor
+public class StorageConnectorService {
+    private final StorageConnectorFactory storageConnectorFactory;
+    private final ObjectMapper objectMapper;
+
+    @SuppressFBWarnings(
+        value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
+        justification = "SpotBugs issue with fix not yet released https://github.com/spotbugs/spotbugs/issues/456")
+    public void handleStructuredRecord(GpcStructuredResponseObject response) throws IOException {
+        String jsonStringResponse = objectMapper.writeValueAsString(response);
```
mattd-kainos, line 28: Found your IOException. It would be quite the exception indeed if this writeValueAsString failed with a JSONProcessingException. You would not expect the serialised of a simple internal object to fail unless there is a real programming error.

I would annotate `handleStructuredRecord` with `@SneakyThrows(JsonProcessingException.class)`. I think in this case suppressing the checked exception can be justified.

e2e-tests/src/main/java/uk/nhs/adaptors/gp2gp/Mongo.java
```diff
@@ -20,4 +21,15 @@ public static Document findEhrExtractStatus(String conversationId) {
         return collection.find(Filters.eq("conversationId", conversationId)).first();
     }
 
+    public static Document findUpdatedEhrExtractStatus(String conversationId, String orginalUpdatedTime) {
```
mattd-kainos, line 24: unused method

service/src/test/java/uk/nhs/adaptors/gp2gp/common/task/TaskConsumerTest.java
```diff
@@ -26,6 +27,9 @@
     @Mock
     private MDCService mdcService;
 
+    @Mock
+    private GpcConfiguration gpcConfiguration;
```
mattd-kainos, line 31: unused

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcPatientDataHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import uk.nhs.adaptors.gp2gp.ehr.EhrExtractStatus;
+import uk.nhs.adaptors.gp2gp.ehr.EhrExtractStatusRepository;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class GpcPatientDataHandler {
```
mattd-kainos, line 20: I think this class name is misleading. It is not actually handling patient data all!

This is our service / repository class for making changes to our ehrExtractStatus database collection.

Maybe we need a top-level package named 'status'? EhrExtractStatus, its repostiory can live there. Rename this `EhrExtractStatusService` which will encapsulate all repository / mongotemplate operation?

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcStructuredAccessException.java
```diff
@@ -0,0 +1,7 @@
+package uk.nhs.adaptors.gp2gp.gpc;
+
+public class GpcStructuredAccessException extends RuntimeException {
```
mattd-kainos, line 3: I had mentioned in another comment throwing a "GpConnectException". Perhaps rename this to "GpConnectException"? 

You'll probably want/need to add the construct passing the root cause - we don't want to throw that away.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcStructuredResponseObject.java
```diff
@@ -0,0 +1,13 @@
+package uk.nhs.adaptors.gp2gp.gpc;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class GpcStructuredResponseObject {
```
mattd-kainos, line 8: All of our storage objects will be wrapped with this exact same metadata. Maybe this should be a `StorageWrapperObject` in the `common.storage` package so it can be re-used?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorService.java
```diff
@@ -0,0 +1,32 @@
...
+import uk.nhs.adaptors.gp2gp.gpc.GpcStructuredResponseObject;
+
+@Service
+@AllArgsConstructor
+public class StorageConnectorService {
+    private final StorageConnector storageConnector;
+    private final ObjectMapper objectMapper;
+
+    @SneakyThrows(JsonProcessingException.class)
+    public void handleStructuredRecord(GpcStructuredResponseObject response) {
```
mattd-kainos, line 25: I had mentioned in a comment below renaming `GpcStructuredResponseObject` to `StorageDataWrapper` since we can use this structure for all storage data. Then `handleStructuredRecord` could be renamed to uploadWithMetadata perhaps?

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcConfiguration.java
```diff
@@ -0,0 +1,17 @@
+package uk.nhs.adaptors.gp2gp.gpc;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Component
+@ConfigurationProperties(prefix = "gp-connect")
```
mattd-kainos, line 10: I'm surprised this prefix works. Javadoc says "A valid prefix is defined by one or more words separated with dots" and the actual value is "gpConnect". Maybe change this to "gpConnect" just to be safe.

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcException.java
```diff
@@ -0,0 +1,8 @@
+package uk.nhs.adaptors.gp2gp.gpc;
+
+public class GpcException extends RuntimeException {
```
mattd-kainos, line 3: I saw another similar RTE somewhere. How about a single type for all GPC-related errors: `GpConnectException`

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -0,0 +1,36 @@
...
+    private final EhrExtractStatusRepository ehrExtractStatusRepository;
+    private final MongoTemplate mongoTemplate;
+
+    public void updateEhrExtractStatusAccessStructured(GetGpcStructuredTaskDefinition structuredTaskDefinition) {
+        var now = now();
+        Query query = new Query();
+        query.addCriteria(Criteria.where("conversationId").is(structuredTaskDefinition.getConversationId()));
+        Update update = new Update();
+        update.set("updatedAt", now);
+        update.set("gpcAccessStructured.$.accessedAt", now);
```
mattd-kainos, line 31: I think you'll just want `gpcAccessStructured.accessedAt` (no $ operator for these)

The `$` operator is for updating arrays, but there is no array in this case. I assume this was carried over from @mmackiewicz1  branch where he does need to use the array update operator?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -0,0 +1,36 @@
...
+        update.set("gpcAccessStructured.$.taskId", structuredTaskDefinition.getTaskId());
+        update.set("gpcAccessStructured.$.objectName", structuredTaskDefinition.getConversationId() + GPC_STRUCTURED_FILE_EXTENSION);
+        mongoTemplate.updateFirst(query, update, EhrExtractStatus.class);
```
mattd-kainos, line 34: This returns an UpdateResult. We expect that exactly one record is modified. Throw an exception otherwise.

mmackiewicz1 (approved):

mattd-kainos (approved):

PR #46, opened 2021-01-19
NIAD-953 Use wiremock by default for local tests

mattd-kainos:
* updated and tested application.yml using wiremock by default
* removed SSP prefixes from wiremock - using SSP is a future story
* updated but not tested docker-compose using wiremock by default


Broom94 (approved):

mattd-kainos:
Moving this back to DRAFT. We will need to update our component tests to use wiremock by default also. I will do that following 788 and 815 merges.

mattd-kainos:
replaced by https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/60

PR #53, opened 2021-01-25
Adaptor documentation updates

mattd-kainos:


mattd-kainos:
never completed and out of date

PR #63, opened 2021-02-02
NIAD-981: Transmission and control act wrappers

msolecka:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/NonEncodedMustacheFactoryAdapter.java
```diff
@@ -0,0 +1,23 @@
+package uk.nhs.adaptors.gp2gp.ehr.utils;
+
+import com.github.mustachejava.DefaultMustacheFactory;
+import com.github.mustachejava.DefaultMustacheVisitor;
+import com.github.mustachejava.MustacheVisitor;
+import com.github.mustachejava.TemplateContext;
+import com.github.mustachejava.codes.ValueCode;
+
+public class NonEncodedMustacheFactoryAdapter extends DefaultMustacheFactory {
```
mattd-kainos, line 9: Was this change to prevent Mustache from escaping HTML/XML characters by default like in this stack overflow? https://stackoverflow.com/questions/22910428/mustache-globally-disable-html-escaping

In most cases this would be the desired behaviour. Most of our further stories have test criteria that these ARE escaped in the XML.

It just the `{{ehrExtractContent}}` we want to avoid escaping, right? We could use triple braces `{{{ehrExtractContent}}}` in the template instead? Disclaimed: I haven't tried it, its just what I've read on the internet! https://www.elated.com/easy-html-templates-with-mustache/

msolecka: hm.. I read on github that they won't provide the option with not escaping and we need to do it by ourselves, but we can  create a specific class like for that purpose and add if statement in the mustache visitor's execute method, e.g. NotEncoded containing the string that will not be encoded.

msolecka: I will check that triple braces, if not prepare the solution with this if in visitor.

mattd-kainos (approved):

z-uddin (approved):

PR #74, opened 2021-02-09
NIAD-1033: Generate ids in the context of message

msolecka:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MapperService.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.stereotype.Component;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import uk.nhs.adaptors.gp2gp.common.service.RandomIdGeneratorService;
+
+@Slf4j
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+@Component
+public class MapperService {
```
mattd-kainos, line 21: I don't understand the purpose of this service. Mapping all narrative statements together is not a GP2GP requirements.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -0,0 +1,20 @@
+package uk.nhs.adaptors.gp2gp.ehr.mapper;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import lombok.AllArgsConstructor;
+import uk.nhs.adaptors.gp2gp.common.service.RandomIdGeneratorService;
+
+@AllArgsConstructor()
+public class IdMapper {
```
mattd-kainos, line 10: Should we change as part of this story that existing mappers use `IdMapper` instead of `RandomIdGeneratorService`?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MapperService.java
msolecka: I need to somewhere initialise message context and clear it after transforming

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -0,0 +1,20 @@
...
+    private final RandomIdGeneratorService randomIdGeneratorService;
+    private final Map<String, String> ids = new HashMap<>();
+
+    public String getOrNew(String id) {
```
mattd-kainos, line 14: I think we'll need our key to be "{resourceType}/{id}" e.g. "Encounter/6CE8318F-2230-47CE-86CC-5210130A00F5". There is no requirement in FHIR that the id is unique across multiple resource types. If the identifier is a number generated sequentially then collisions will be commonplace. http://hl7.org/fhir/stu3/datatypes.html#id

Maybe the method could be something like `public String createReplacementIdentifier(Resource resource)`?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MessageContext.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class MessageContext {
+
+    private static ThreadLocal<IdMapper> idMapperHolder = new NamedThreadLocal<>("IdMapper");
+
+    public static void resetMessageContext() {
+        idMapperHolder.remove();
+    }
+
+    public static void setMessageContext(IdMapper idMapper) {
```
mattd-kainos, line 13: This is only ever called with new IdMapper().

Would it make sense to have an `initialiseMessageContext()` that hides creation of the IdMapper? That way we could add more threadlocals to the this class to capture different parts of the message context

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MessageContext.java
```diff
@@ -0,0 +1,24 @@
+package uk.nhs.adaptors.gp2gp.ehr.mapper;
+
+import org.springframework.core.NamedThreadLocal;
+
+public class MessageContext {
```
mattd-kainos, line 5: With initialiseMessageContext this could be a `@Component` and autowired

mattd-kainos (changes requested):
Using a FHIR reference for our key is needed to avoid issues in the future. Suggestions for changing method names, converting to beans etc. optional. Question about the MapperService. I'm not sure this is needed.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
msolecka: Yes, NarrativeStatementMapper was updated to use IdMapper, Do we have more mappers?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
msolecka: I think the method could get Fhir Reference as parameter because the reference keeps the resource type and the id and while providing mapping we will be mapping more than id of the resource but also the references it has inside. Passing full resource to generate all ids for it, would move much logic that will be already in mapping to find all the references inside and I don't think it should be the responsibility of that class.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MessageContext.java
msolecka: good idea. I noticed I could also create the mapper when fetching it for the first time.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -0,0 +1,32 @@
...
+@AllArgsConstructor()
+public class IdMapper {
+    private final RandomIdGeneratorService randomIdGeneratorService;
+    private final Map<String, String> ids = new HashMap<>();
+
+    public String getOrNew(ResourceType resourceType, String id) {
+        return getOrNew(buildReference(resourceType, id));
+    }
+
+    public String getOrNew(Reference reference) {
```
mattd-kainos, line 22: Nice! we need both of these.

mattd-kainos (approved):

mmackiewicz1 (approved):

Broom94 (approved):

PR #82, opened 2021-02-12
Feature/niad 1040 output ehr composition for each encounter

mmackiewicz1:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+import org.hl7.fhir.dstu3.model.Bundle;
+import org.hl7.fhir.dstu3.model.Encounter;
+import org.hl7.fhir.dstu3.model.ListResource;
+import org.hl7.fhir.dstu3.model.Reference;
+import org.hl7.fhir.dstu3.model.ResourceType;
+
+public final class EncounterExtractor {
+    private static final String CONSULTATION_LIST_CODE = "1149501000000101";
+
+    private EncounterExtractor() {
```
msolecka, line 16: Is this constructor needed?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+            .filter(entry -> entry.getResource().getResourceType().equals(ResourceType.Encounter))
+            .map(entry -> (Encounter) entry.getResource())
+            .filter(encounter -> encounterReferences.contains(encounter.getId()))
+            .collect(Collectors.toList());
+    }
+
+    private static boolean isConsultationList(ListResource listResource) {
+        return listResource.getCode()
+            .getCoding()
+            .get(0)
```
msolecka, line 41: Are we sure about single code only? I would check if any of the codes match the consultation code.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
mmackiewicz1: It prevents creating an instance of this class. It is for static use only.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
```diff
@@ -57,11 +64,15 @@ private EhrExtractTemplateParameters prepareEhrFhirExtractParamsFromFhirBundle(
         ehrExtractTemplateParameters.setToOdsCode(getGpcStructuredTaskDefinition.getToOdsCode());
         ehrExtractTemplateParameters.setFromOdsCode(getGpcStructuredTaskDefinition.getFromOdsCode());
         ehrExtractTemplateParameters.setAvailabilityTime(DATE_TIME_FORMATTER.format(timestampService.now()
-            .atOffset(ZoneOffset.UTC)));
+            .atZone(ZoneId.of(UK_ZONE_ID))));
 
         extractPatientFromBundle(bundle)
             .orElseThrow(() -> new FhirValidationException("Missing patient resource in Fhir Bundle."));
 
+        var encounters = EncounterExtractor.extractEncountersByReferences(bundle.getEntry(),
```
msolecka, line 72: For me it would be easier understand the business logic that' behind list's mapping, if we would have just one method  extractEncountersFromConsultationsList(bundle.getEntry()) , instead of two.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+import org.hl7.fhir.dstu3.model.Reference;
+import org.hl7.fhir.dstu3.model.ResourceType;
+
+public final class EncounterExtractor {
+    private static final String CONSULTATION_LIST_CODE = "1149501000000101";
+
+    private EncounterExtractor() {
+    }
+
+    public static List<String> extractEncounterReferences(List<Bundle.BundleEntryComponent> entries) {
```
msolecka, line 19: maybe extractEncounterReferencesFromConsultationList?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
mmackiewicz1: I was thinking about this. However, notice that EncounterExtractor is a class which only extracts encounters from bundle, while the mapper class maps provided encounter to an xml format. I think we should think here about single responsibility principle.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.hl7.fhir.dstu3.model.Bundle;
+import org.hl7.fhir.dstu3.model.Encounter;
+import org.hl7.fhir.dstu3.model.ListResource;
+import org.hl7.fhir.dstu3.model.Reference;
+import org.hl7.fhir.dstu3.model.ResourceType;
+
+public final class EncounterExtractor {
```
msolecka, line 13: It would be good to add tests for that class.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
msolecka: EncounterExtractor has method extractEncountersByReferences that returns list of Encounters.

msolecka: I was thinking about moving logic of extractEncountersByReferences and extractEncountersByReferences into one, while they are both from the same class.

mmackiewicz1: Ok

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
msolecka: right, I didn't notice it is private, what about @NoArgsConstructor(access = AccessLevel.PRIVATE) from lombok ?

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapperTest.java
```diff
@@ -80,11 +85,20 @@ public static void initialize() throws IOException {
...
     }
 
+    @AfterAll
+    public static void deinitialize() {
+        TimeZone.setDefault(null);
+    }
+
     @BeforeEach
     public void setUp() {
         when(randomIdGeneratorService.createNewId()).thenReturn(TEST_ID);
```
msolecka, line 95: the test could return the list of values to verify if there aren't same records on the output.

msolecka: or, you can pre-populate the messageContext with correct id - it seems to be easier to maintain and verify your output xml is valid. Also if the input message contained less Encounters references, the output xml would be easier to test and maintain in the future. I also think it would be valuable to add test for empty list and when the list is not in the bundle.

docker/wiremock/stubs/__files/correctPatientStructuredRecordResponse.json
```diff
@@ -1 +1 @@
...
...ibingAgency-1","code":"prescribed-at-gp-practice","display":"Prescribed at GP practice"}]}},{"url":"https://fhir.nhs.uk/STU3/StructureDefinition/Extension-CareConnect-GPC-MedicationStatementLastIssueDate-1","valueDateTime":"2010-10-20T00:00:00+01:00"}],"identifier":[{"system":"https://EMISWeb/A82038","value":"57E4E63C-DEFE-B7BB-2652-4BB20FA31362"}],"basedOn":[{"reference":"MedicationRequest/47784A83-5675-4EF2-9947-0D7D125F1656-FD-1"}],"status":"completed","medicationReference":{"reference":"Medication/9CA00C4D-E545-4A18-BA84-C74D74DD7E95"},"effectivePeriod":{"start":"2010-01-13","end":"2010-11-17"},"dateAsserted":"2010-01-13T11:37:34.423+00:00","subject":{"reference":"Patient/88F14BF6-CADE-47D6-90E2-B10519BF956F"},"taken":"unk","dosage":[{"text":"One Tablet Daily For 21 Days; Subsequent Courses Repeated After 7-Day Tablet Free Interval"}]}},{"resource":{"resourceType":"MedicationStatement","id":"581EC356-84CA-450B-9D0A-DADEAAD743D3-FD-1-MS","meta":{"profile":["https://fhir.nhs.uk/STU3/StructureDefinition/CareConnect-GPC-MedicationStatement-1"]},"extension":[{"url":"https://fhir.nhs.uk/STU3/StructureDefinition/Extension-CareConnect-GPC-PrescribingAgency-1","valueCodeableConcept":{"c
```
mattd-kainos, line 1: We intend for our wiremock to mirror the patients that are in the GP connect demonstrator.

Instead of replacing the data that is already there, could you add this as an additional patient? I assume it is a test patient from one of the example datasets. Could you make the NHS number for the wiremock patient the same as the NHS number for the example dataset?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrDocumentMapper.java
```diff
@@ -25,14 +25,15 @@
...
+    private static final String UK_ZONE_ID = "Europe/London";
 
     private final TimestampService timestampService;
     private final RandomIdGeneratorService randomIdGeneratorService;
 
     public EhrDocumentTemplateParameters mapToMhsPayloadTemplateParameters(GetGpcDocumentTaskDefinition taskDefinition, String messageId) {
         return EhrDocumentTemplateParameters.builder()
             .resourceCreated(DATE_TIME_FORMATTER.format(timestampService.now()
-                .atOffset(ZoneOffset.UTC)))
+                .atZone(ZoneId.of(UK_ZONE_ID))))
```
mattd-kainos, line 36: Use DateFormatUtil instead?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
```diff
@@ -57,11 +64,14 @@ private EhrExtractTemplateParameters prepareEhrFhirExtractParamsFromFhirBundle(
         ehrExtractTemplateParameters.setToOdsCode(getGpcStructuredTaskDefinition.getToOdsCode());
         ehrExtractTemplateParameters.setFromOdsCode(getGpcStructuredTaskDefinition.getFromOdsCode());
         ehrExtractTemplateParameters.setAvailabilityTime(DATE_TIME_FORMATTER.format(timestampService.now()
-            .atOffset(ZoneOffset.UTC)));
+            .atZone(ZoneId.of(UK_ZONE_ID))));
```
mattd-kainos, line 67: Use DateFormatUtil instead?

docker/wiremock/stubs/__files/correctPatientStructuredRecordResponse.json
mmackiewicz1: Would it be enough to add "Encounter" and "List" resources?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/OutputMessageWrapperMapper.java
```diff
@@ -24,17 +25,18 @@
...
 
     private final RandomIdGeneratorService randomIdGeneratorService;
     private final TimestampService timestampService;
 
-    public String map(GetGpcStructuredTaskDefinition getGpcDocumentTaskDefinition, String content) {
+    public String map(GetGpcStructuredTaskDefinition getGpcDocumentTaskDefinition, String ehrExtractContent) {
         OutputMessageWrapperTemplateParameters outputMessageWrapperTemplateParameters = OutputMessageWrapperTemplateParameters.builder()
             .eventId(randomIdGeneratorService.createNewId())
-            .creationTime(DATE_TIME_FORMATTER.format(timestampService.now().atOffset(ZoneOffset.UTC)))
+            .creationTime(DATE_TIME_FORMATTER.format(timestampService.now().atZone(ZoneId.of(UK_ZONE_ID))))
```
mattd-kainos, line 36: use DateFormatUtil instead?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrDocumentMapper.java
mmackiewicz1: It does not support "Instant" types. Shall I convert them to Date time?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
```diff
@@ -0,0 +1,54 @@
...
+import org.hl7.fhir.dstu3.model.Reference;
+import org.hl7.fhir.dstu3.model.ResourceType;
+import org.springframework.stereotype.Component;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@Component
+public final class EncounterExtractor {
+    private static final String CONSULTATION_LIST_CODE = "1149501000000101";
+
+    public static List<Encounter> extractConsultationListEncounters(List<Bundle.BundleEntryComponent> entries) {
```
mattd-kainos, line 22: Are the encounters in the returned `List<Encounter>` in the same order at they appear in the Encounter List? I'd like to assert this is the case.

mmackiewicz1: Yes, their order is being asserted in a test. Order is set according to references order in the consultation list.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapperTest.java
```diff
@@ -80,11 +89,26 @@ public static void initialize() throws IOException {
             .build();
     }
 
+    @AfterAll
+    public static void deinitialize() {
+        TimeZone.setDefault(null);
```
mattd-kainos, line 94: Setting/resetting the default timezone should not be required. Either remove here, or ensure that https://gpitbjss.atlassian.net/browse/NIAD-1060 removes them

mmackiewicz1: Removal of these methods will cause different timings to be asserted when run in different time zones.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapperTest.java
```diff
@@ -80,11 +89,26 @@ public static void initialize() throws IOException {
...
+        messageContext = new MessageContext(randomIdGeneratorService);
+        ehrExtractMapper = new EhrExtractMapper(new FhirParseService(),
+            randomIdGeneratorService,
+            timestampService,
+            new EncounterMapper(messageContext));
+    }
+
+    @AfterEach
+    public void tearDown() {
+        reset();
```
mattd-kainos, line 110: you shouldn't ever need to manually reset your Mockito mocks

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ImmunizationObservationStatementMapperTest.java
```diff
@@ -46,6 +51,16 @@
     private Bundle bundle;
     private FhirParseService fhirParseService;
 
+    @BeforeAll
+    public static void initialize() {
+        TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC));
```
mattd-kainos, line 56: as above, remove or ensure @Broom94 is aware that https://gpitbjss.atlassian.net/browse/NIAD-1060 needs to remove it

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ImmunizationObservationStatementMapperTest.java
```diff
@@ -58,14 +73,10 @@ public void setUp() throws IOException {
 
     @AfterEach
     public void tearDown() {
+        reset();
```
mattd-kainos, line 76: as above about resetting mocks

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/NarrativeStatementMapperTest.java
```diff
@@ -53,6 +57,11 @@ public static void initialize() {
         TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC));
     }
 
+    @AfterAll
+    public static void deinitialize() {
+        TimeZone.setDefault(null);
```
mattd-kainos, line 62: remove or @Broom94 / https://gpitbjss.atlassian.net/browse/NIAD-1060

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/NarrativeStatementMapperTest.java
```diff
@@ -62,14 +71,10 @@ public void setUp() {
 
     @AfterEach
     public void tearDown() {
+        reset();
```
mattd-kainos, line 74: shouldn't use reset

mattd-kainos (changes requested):
good work getting packages and classes renamed consistently
just a few points
- Ensure the compositions are output in the same order they appear in the encounter list. This is very important.
- Mockito.reset should generally never be used
- Expanding use of TimeZone.setDefault either needs to be avoiding, or mitigated with agreement that NIAD-1060 will remove all of these statements

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrDocumentMapper.java
mattd-kainos: Add a method for producing the string from instant

docker/wiremock/stubs/__files/correctPatientStructuredRecordResponse.json
mattd-kainos: I would recommend creating an ASR with just the encounter list and encounter resources for the unit test. We should use complete patient records for the wiremock.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
mattd-kainos: The test works because the bundle happens to have the encounters in the same order as the encounter list.

I create a branch where I swapped the order of two encounters in the bundle to make the test fail. https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/87/files

We can go through the code tomorrow to look at why if you want.

mmackiewicz1:
Removed time zoning waiting for the upcoming fix.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
mattd-kainos: extractEncounterReferencesFromEncounterList would be a better name

Even though the display is "List of consultations" there is yet another type of list we'll see in NIAD-904 named "Consultation" which is referred to as the List(Consultation) https://developer.nhs.uk/apis/gpconnect-1-5-0/accessrecord_structured_development_list_consultation.html

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EncounterExtractor.java
mattd-kainos: Also protect against the missing code/coding cases

msolecka (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -23,4 +24,14 @@ public static String formatDate(Date date) {
                 .atZone(ZoneId.of(UK_ZONE_ID))
                 .toLocalDateTime());
     }
+
+    public static String formatDate(Instant instant) {
+        if (instant == null) {
```
Broom94, line 29: I don't think we should pass null here, you could also use getValue

service/src/intTest/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredComponentTest.java
```diff
@@ -37,24 +27,40 @@
 import uk.nhs.adaptors.gp2gp.testcontainers.ActiveMQExtension;
 import uk.nhs.adaptors.gp2gp.testcontainers.MongoDBExtension;
 
+import org.apache.commons.io.IOUtils;
```
Broom94, line 30: Do you have the correct code config?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ImmunizationObservationStatementMapper.java
```diff
@@ -156,7 +148,8 @@ private String buildLotNumberPertinentInformation(Immunization immunization) {
 
     private String buildExpirationDatePertinentInformation(Immunization immunization) {
         Optional<Date> expirationDate = Optional.ofNullable(immunization.getExpirationDate());
-        return expirationDate.map(value -> EXPIRATION + formatShortDate(value)).orElse(StringUtils.EMPTY);
+        return expirationDate.map(date -> EXPIRATION + DateFormatUtil.formatDate(date))
+            .orElse(StringUtils.EMPTY);
```
Broom94, line 152: static import EMPTY?

Broom94 (changes requested):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ImmunizationObservationStatementMapper.java
mmackiewicz1: It is better to have a name of the class in front for better readability. Simple "EMPTY" does not say much.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredComponentTest.java
mmackiewicz1: The same as in the res of the classes.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
mmackiewicz1: Instant does not have "getValue" method. Do you mean something else?

mattd-kainos (approved):

mattd-kainos (approved):

msolecka (approved):

PR #87, opened 2021-02-15
NIAD-1040 Test to show output is in order of bundle not encounter list

mattd-kainos:
Changing order of Encounters within the bundle causes ehrComposition to be output in the wrong order


mattd-kainos:
no longer needed

PR #92, opened 2021-02-17
NIAD-1060: FHIR to HL7 date and times

Broom94:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,48 @@
...
+                return localDateTime(dateTimeType, "yyyyMM");
+            case DAY:
+                return localDateTime(dateTimeType, "yyyyMMdd");
+            default:
+                return localDateTime(dateTimeType, "yyyyMMddHHmmss");
+        }
+    }
+
+    private static String localDateTime(DateTimeType effectiveDateTimeType, String pattern) {
+        return LocalDateTime.ofInstant(effectiveDateTimeType.toCalendar().toInstant(), ZoneId.of("Europe/London"))
```
z-uddin, line 35: Could change this back to use a variable?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,48 @@
...
+    public static String formatDateTimeType(DateTimeType dateTimeType) {
+        if (dateTimeType == null) {
             throw new EhrMapperException("Could not format date");
         }
 
-        return DATE_TIME_FORMATTER.format(
-            date.toInstant()
-                .atZone(ZoneId.of(UK_ZONE_ID))
-                .toLocalDateTime());
+        switch (dateTimeType.getPrecision()) {
```
mmackiewicz1, line 22: Is it possible to use a HashMap here?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,48 @@
...
-    private static final String UK_ZONE_ID = "Europe/London";
-    private static final DateTimeFormatter DATE_TIME_FORMATTER = new DateTimeFormatterBuilder()
-        .appendPattern("yyyyMMddHHmmss")
-        .toFormatter();
 
-    public static String formatDate(Date date) {
-        if (date == null) {
+    @SneakyThrows
+    public static String formatDateTimeType(DateTimeType dateTimeType) {
+        if (dateTimeType == null) {
```
mmackiewicz1, line 18: I think that the null should be checked be outside of the method. Passing null values (most of the time) is a bad practice.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
mmackiewicz1, line 35: In my PR I created a date time utils method for instants. Maybe you could copy this change and use it here?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
msolecka: I don't agree, it is good that class gives a meaningful error instead of null pointer exception. This null will come here only in case of error.

mmackiewicz1: We can throw this exception outside of method, right before it is invoked.

msolecka: but then everybody who will ever use that public method with have to know about it.

mmackiewicz1: @Broom94 , are we planning to pass a null value here often?

ElliotB-96 (approved):
Looks good

z-uddin (approved):

z-uddin (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ImmunizationObservationStatementMapper.java
```diff
@@ -149,8 +150,16 @@ private String buildLotNumberPertinentInformation(Immunization immunization) {
...
+        Optional<DateType> expirationDateElement = Optional.ofNullable(immunization.getExpirationDateElement());
+        if (expirationDateElement.isPresent()) {
+            if (expirationDateElement.get().hasValue()) {
+                return expirationDateElement.map(dateType -> EXPIRATION + formatDateType(dateType)).orElse(StringUtils.EMPTY);
+            } else {
+                return StringUtils.EMPTY;
+            }
+        } else {
+            return StringUtils.EMPTY;
+        }
```
msolecka, line 162:
```suggestion
        Optional<DateType> expirationDateElement = Optional.ofNullable(immunization.getExpirationDateElement());
        if (expirationDateElement.isPresent() && expirationDateElement.get().hasValue()) {
            return expirationDateElement.map(dateType -> EXPIRATION + formatDateType(dateType)).orElse(StringUtils.EMPTY);
        } 
           
        return StringUtils.EMPTY;
```

msolecka: There could be only one if and one StringUtils.EMPTY; to achieve the same effect.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,73 @@
...
+    public static final String COULD_NOT_FORMAT_DATE = "Could not format date";
+    private static final ImmutableMap<TemporalPrecisionEnum, String> YEAR_PATTERN_MAP = ImmutableMap.of(
+        YEAR, "yyyy",
+        MONTH, "yyyyMM",
+        DAY, "yyyyMMdd",
+        MINUTE, "yyyyMMddHHmm"
+    );
+
+    public static String formatDateType(DateType dateType) {
+        if (!dateType.hasValue()) {
```
msolecka, line 35: 👍

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,73 @@
...
+            throw new EhrMapperException(COULD_NOT_FORMAT_DATE);
+        }
+
+        return Optional.ofNullable(YEAR_PATTERN_MAP.get(dateType.getPrecision()))
+            .map(pattern -> dateTypeConverter(dateType, pattern))
+            .orElseGet(() -> dateTypeConverter(dateType, "yyyyMMddHHmmss"));
```
msolecka, line 41: the default value could be const as it is used in more than one place.

service/src/test/resources/ehr/mapper/immunization/expected-output-observation-statement-all-information.xml
```diff
@@ -9,7 +9,7 @@
...
... a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is a long paragraph. This is the END of the long paragraph. Reason: Immunisation due Vaccination Protocol 1: A vaccination for Hepatitis
```
msolecka, line 12: I am not sure about changing the date format in the text is what we should do. In the story I am working on now, it has been told to format date in text in different way - more user friendly - than in effectiveTime or availabilityTime. I think yyyy-MM-dd 2011-06-21 format should stay in this case.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,73 @@
...
+
+import org.hl7.fhir.dstu3.model.DateTimeType;
+import org.hl7.fhir.dstu3.model.DateType;
+import org.hl7.fhir.dstu3.model.InstantType;
+
+import com.google.common.collect.ImmutableMap;
 
 import uk.nhs.adaptors.gp2gp.ehr.exception.EhrMapperException;
 
 public class DateFormatUtil {
```
mattd-kainos, line 23: We should have unit tests for this to cover various boundary and edge cases. This will be difficult to verify otherwise. A few ideas:

* Dates/times in both BST and GMT
* Differing time zone offsets in the FHIR/JSON representation

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,73 @@
...
-        .toFormatter();
-
-    public static String formatDate(Date date) {
-        if (date == null) {
-            throw new EhrMapperException("Could not format date");
+
+    public static final String UK_TIMEZONE = "Europe/London";
+    public static final String COULD_NOT_FORMAT_DATE = "Could not format date";
+    private static final ImmutableMap<TemporalPrecisionEnum, String> YEAR_PATTERN_MAP = ImmutableMap.of(
+        YEAR, "yyyy",
```
mattd-kainos, line 28: since DateTimeFormatter is threadsafe and immutable you could consider creating the formatter here and reusing it instead of creating a new formatter from its string representation every time.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/DateFormatUtil.java
```diff
@@ -1,26 +1,73 @@
...
+        if (!dateTimeType.hasValue()) {
+            throw new EhrMapperException(COULD_NOT_FORMAT_DATE);
+        }
+
+        return Optional.ofNullable(YEAR_PATTERN_MAP.get(dateTimeType.getPrecision()))
+            .map(pattern -> dateTimeTypeConverter(dateTimeType, pattern))
+            .orElseGet(() -> dateTimeTypeConverter(dateTimeType, "yyyyMMddHHmmss"));
+    }
+
+    private static String dateTypeConverter(DateType effectiveDateTimeType, String pattern) {
```
mattd-kainos, line 64: Consider replacing these two methods with this:

```
    private static String getFormatStringForPrecision(TemporalPrecisionEnum precisionEnum) {
        return YEAR_PATTERN_MAP.getOrDefault(precisionEnum, "yyyyMMddHHmmss");
    }

    private static String convertWithPrecision(TemporalPrecisionEnum precisionEnum, Calendar calendar) {
        return LocalDateTime.ofInstant(calendar.toInstant(), ZoneId.of(UK_TIMEZONE))
            .format(DateTimeFormatter.ofPattern(getFormatStringForPrecision(precisionEnum)));
    }
```

Then each FHIR-type specific method is only concerned with getting the Calendar object from HAPI's *Type. The implementation would be slightly different if you've already created the formatters as suggested above.

mattd-kainos (changes requested):
Util class needs unit test coverage.

mmackiewicz1 (approved):

service/src/test/java/uk/nhs/adaptors/gp2gp/utils/DateFormatUtilTest.java
```diff
@@ -0,0 +1,59 @@
...
+//            Arguments.of(format(obsInstantDateTimeTemplate, "2017-02-28", "2017-02-28"),
+//                format(immDateTypeTemplate, "2017-02-28"), "20170228"),
+//            Arguments.of(format(obsInstantDateTimeTemplate, "2019-01-28T10:30", "2019-01-28T10:30"),
+//                format(immDateTypeTemplate, "2019-01-28T10:30"), "201901281030"),
+//            Arguments.of(format(obsInstantDateTimeTemplate, "2010-07-28T10:30:00", "2019-07-28T10:30:00"),
+//                format(immDateTypeTemplate, "2019-07-28T10:30:00"), "20190728103000"),
+//            Arguments.of(format(obsInstantDateTimeTemplate, "2019-03-28T10:30:00+00:00", "2019-03-28T10:30:00+00:00"),
+//                format(immDateTypeTemplate, "2019-07-28T10:30:00+01:00"), "20190728103000")
+//        );
+//    }
```
ElliotB-96, line 58: I take it these tests are no longer being used?

mmackiewicz1 (approved):

service/src/test/java/uk/nhs/adaptors/gp2gp/utils/DateFormatUtilTest.java
```diff
@@ -0,0 +1,73 @@
...
+        String observationJson = String.format(INSTANT_OBSERVATION_TEMPLATE, input);
+        Observation observation = FHIR_PARSER.parseResource(observationJson, Observation.class);
+
+        String actual = toHl7Format(observation.getIssuedElement());
+        assertThat(actual).isEqualTo(expected);
+    }
+
+    @ParameterizedTest
+    @MethodSource("dateParams")
+    public void When_FormattingDateTypeToHl7_Expect_Hl7InUkZone(String input, String expected) {
```
mattd-kainos, line 37: This is testing InstantType not DateType. It appears that "valueDate" (DateType) is not valid in Observation (or any resource used by GP Connect) so this test can be removed. Don't forgot to remove the unused template.

ElliotB-96 (dismissed):

service/src/test/java/uk/nhs/adaptors/gp2gp/utils/DateFormatUtilTest.java
mattd-kainos: Use a different resource type that does support dateType

mattd-kainos (approved):

ElliotB-96 (approved):

PR #96, opened 2021-02-18
NIAD-916 Translating Referral Request to Request Statement

z-uddin:
* Created a RequestStatement mapper, which consumes FHIR Referral Request object and uses this to fill a Request Statement XML template 
* Unit tests parse a JSON file to a FHIR Referral Request object and then compares output from running map method to expected XML


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,296 @@
...
+                    .map(value -> REQUESTER_RELATION + getHumanNameString(value.getNameFirstRep()))
+                    .orElse(StringUtils.EMPTY);
+            }
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private String getDeviceCode(Device device) {
+        return extractTextOrCoding(device.getType()).orElse(StringUtils.EMPTY);
```
mmackiewicz1, line 165: I would put "orElse" in a new line for a better readability.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,296 @@
...
+                .filter(Reference::hasReferenceElement)
+                .map(value -> extractRecipientString(value))
+                .collect(Collectors.joining(COMMA));
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private List<Reference> removeFirstPractitionerReference(List<Reference> referenceList) {
+        for (int i = 0; i < referenceList.size() - 1; i++) {
```
mmackiewicz1, line 191: Maybe we could name the "i" variable "index"?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,296 @@
...
+    private String buildNoteDescription(ReferralRequest referralRequest) {
+        if (referralRequest.hasNote()) {
+            return referralRequest.getNote().stream()
+                .map(value -> String.format(NOTE, getAuthorString(value), getNoteTime(value),  value.getText()))
+                .collect(Collectors.joining(COMMA));
+        }
+        return StringUtils.EMPTY;
+    }
+
+    private String getAuthorString(Annotation annotation) {
```
mmackiewicz1, line 249:
```suggestion
    private String extractAuthor(Annotation annotation) {
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,296 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import uk.nhs.adaptors.gp2gp.ehr.utils.DateFormatUtil;
+import uk.nhs.adaptors.gp2gp.ehr.utils.StatementTimeMappingUtils;
+import uk.nhs.adaptors.gp2gp.ehr.utils.TemplateUtils;
+
+@Component
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class RequestStatementMapper {
```
mmackiewicz1, line 36: This class is almost 300 lines long. Maybe we could think about moving some of this code to another class? Like "RequestStatementExtractor"?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,296 @@
...
+        for (int i = 0; i < referenceList.size() - 1; i++) {
+            if (referenceList.get(i).getReference().startsWith(ResourceType.Practitioner.name())) {
+                referenceList.remove(i);
+                return referenceList;
+            }
+        }
+        return referenceList;
+    }
+
+    private String extractRecipientString(Reference reference) {
```
mmackiewicz1, line 200: Phrase "String" is redundant.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private String getHumanNameString(HumanName humanName) {
+        if  (humanName.hasText()) {
+            return humanName.getText();
+        } else if (humanName.hasGiven() || humanName.hasFamily()) {
+            String given = Optional.ofNullable(humanName.getGivenAsSingleString()).orElse(StringUtils.EMPTY);
```
mmackiewicz1, line 278: Wrapping variables in optionals for the sake of using "orElse" is not something which we should do. How about using a string builder and append text if present?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/StatementTimeMappingUtils.java
```diff
@@ -38,6 +39,14 @@ public static String prepareAvailabilityTimeForEncounter(Encounter encounter) {
         }
     }
 
+    public static String prepareAvailabilityTimeForReferralRequest(ReferralRequest referralRequest) {
+        if (referralRequest.hasAuthoredOn()) {
+            return String.format(AVAILABILITY_TIME_VALUE_TEMPLATE, DateFormatUtil.formatDate(referralRequest.getAuthoredOn()));
+        } else {
```
mmackiewicz1, line 45: "else" could be omitted;

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapperTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import uk.nhs.adaptors.gp2gp.common.service.FhirParseService;
+import uk.nhs.adaptors.gp2gp.common.service.RandomIdGeneratorService;
+import uk.nhs.adaptors.gp2gp.utils.ResourceTestFileUtils;
+
+@ExtendWith(MockitoExtension.class)
+public class RequestStatementMapperTest {
+    private static final String TEST_ID = "394559384658936";
+    private static final String TEST_FILE_DIRECTORY = "/ehr/mapper/referral/";
+    private static final String INPUT_JSON_BUNDLE =  TEST_FILE_DIRECTORY + "fhir-bundle.json";
```
mmackiewicz1, line 33: Good use of string concatenation :)

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapperTest.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @AfterEach
+    public void tearDown() {
+        messageContext.resetMessageContext();
+    }
+
+    @AfterAll
+    public static void deinitialize() {
+        TimeZone.setDefault(null);
```
mmackiewicz1, line 96: We no longer have to use time zoning.

z-uddin: NIAD-1060 has not been merged into main yet, so will keep the timezone code for the unit tests for now.

mmackiewicz1: Ok

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,202 @@
...
+                .filter(val -> val.getSystem().equals(UBRN_SYSTEM_URL))
+                .map(Identifier::getValue)
+                .findFirst();
+        }
+        return identifier.map(value -> UBRN + value).orElse(StringUtils.EMPTY);
+    }
+
+    private String buildPriorityDescription(ReferralRequest referralRequest) {
+        if (referralRequest.hasPriority()) {
+            return PRIORITY + referralRequest.getPriority().getDisplay();
```
mattd-kainos, line 108: This will most likely need to use the CodeableConceptMappingUtils::extractTextOrCoding "display or coding[0].display" but worth confirming

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,202 @@
...
+            return ignoreFirstPractitioner.stream()
+                .filter(Reference::hasReferenceElement)
+                .map(value -> extractRecipient(messageContext, value))
+                .collect(Collectors.joining(COMMA));
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private List<Reference> removeFirstPractitionerReference(List<Reference> referenceList) {
```
mattd-kainos, line 172: Don't throw it out, we'll need it later! ;) 

How about
```
referralRequest.getRecipient()
  .stream()
  .skip(1)
```
instead?

mattd-kainos: Oh its more complicated than that isn't it. We want to skip the first occurrence of a specific resource.

In that case, create a new list from `referralRequest.getRecipient()` but don't modify the original HAPI FHIR objects. We don't want any data going missing.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
```diff
@@ -0,0 +1,128 @@
...
+                .getResource(referenceId)
+                .map(resource -> (Organization) resource)
+                .map(value -> RECIPIENT_ORG + value.getName())
+                .orElse(StringUtils.EMPTY);
+        }
+        return StringUtils.EMPTY;
+    }
+
+    public static String extractReasonCode(ReferralRequest referralRequest) {
+        var ignoreFirstReasonCode = referralRequest.getReasonCode().subList(1, referralRequest.getReasonCode().size());
```
mattd-kainos, line 72: Could sublist throw ArrayIndexOutOfBounds if there are no reason codes?

It would be safer to use skip(1) on the stream.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
```diff
@@ -0,0 +1,128 @@
...
+        if (annotation.hasAuthorStringType() && annotation.getAuthorStringType().hasValue()) {
+            return NOTE_AUTHOR + annotation.getAuthorStringType().getValue();
+        } else if (annotation.hasAuthorReference() && annotation.getAuthorReference().hasReferenceElement()) {
+            IIdType reference = annotation.getAuthorReference().getReferenceElement();
+            if (reference.getResourceType().equals(ResourceType.RelatedPerson.name())) {
+                return messageContext.getInputBundleHolder()
+                    .getResource(reference)
+                    .map(resource -> (RelatedPerson) resource)
+                    .map(value -> NOTE_AUTHOR_RELATION + extractHumanName(value.getNameFirstRep()))
+                    .orElse(StringUtils.EMPTY);
```
mattd-kainos, line 90: Confirm correct behaviour if we cannot resolve the reference (current impl is to leave it blank)

mattd-kainos: Also confirm correct behaviour if the reference is anything other than RelatedPerson, Practitioner, or Patient

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,202 @@
...
+        if (referralRequest.hasServiceRequested()) {
+            return SERVICE_REQUESTED + extractServiceRequested(referralRequest);
+        }
+        return StringUtils.EMPTY;
+    }
+
+    private String buildRequesterDescription(ReferralRequest referralRequest) {
+        if (referralRequest.hasRequester() && referralRequest.getRequester().getAgent().hasReference()) {
+            IIdType reference = referralRequest.getRequester().getAgent().getReferenceElement();
+            if (reference.getResourceType().equals(ResourceType.Device.name())) {
```
mattd-kainos, line 123: As below, confirm correct behaviour is requester reference cannot be resolved or is an expected type. Do we throw an error and stop processing or is it ok to continue with a blank as done here.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapperTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void When_MappingReferralRequestJsonWithNestedTrue_Expect_RequestStatementXmlOutput() throws IOException {
+        expectedOutputMessage = ResourceTestFileUtils.getFileContent(OUTPUT_XML_USES_NESTED_COMPONENT);
+        var jsonInput = ResourceTestFileUtils.getFileContent(INPUT_JSON_WITH_NO_OPTIONAL_FIELDS);
+        ReferralRequest parsedReferralRequest = new FhirParseService().parseResource(jsonInput, ReferralRequest.class);
+
+        String outputMessage = requestStatementMapper.mapReferralRequestToRequestStatement(parsedReferralRequest, true);
+
+        assertThat(outputMessage).isEqualToIgnoringWhitespace(expectedOutputMessage);
```
mattd-kainos, line 123: We can't use isEqualToIgnoringWhitespace here.

In your XML fixtures each "item" in the <text> element is on a newline. You actual output (should) correctly use a space delimiter for each item, but this test doesn't assert that? 

If, in the future, someone accidentally changed the code to use a newline, tab etc delimiter this test would not detect the regression.

mattd-kainos (changes requested):
Clean code! Two main issues 1) modifying the FHIR objects is a bad idea 2) test might not catch regressions

Also a few things to confirm about wether we can use blanks or need to throw and exception.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
z-uddin: From the HAPI FHIR ReferralRequest class, ReferralPriority is an Enum and not a Codeable concept.
Therefore we cannot use the CodeableConceptMappingUtils.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
z-uddin: Confirmed that if a resource is not resolved an exception should be thrown. 
And that if a reference is not of expected resource type then an exception should also be thrown.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
z-uddin: Confirmed that if a resource is not resolved an exception should be thrown. 
And that if a reference is not of expected resource type then an exception should also be thrown.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    private void removeFirstPractitionerReference(List<Reference> referenceList) {
+        for (int index = 0; index < referenceList.size() - 1; index++) {
+            if (referenceList.get(index).getReference().startsWith(ResourceType.Practitioner.name())) {
+                referenceList.remove(index);
+                return;
+            }
+        }
+    }
```
mattd-kainos, line 184: This code still removes an item from the underlying referralRequest.getRecipient() list. You need to create an entirely new lines that only contains the objects you need. One way to do this is by copying just those objects to the new list.

```suggestion
private String buildRecipientDescription(ReferralRequest referralRequest) {
        if (referralRequest.hasRecipient()) {
            return getRecipientsWithoutFirstPractitioner(referralRequest).stream()
                .filter(Reference::hasReferenceElement)
                .map(value -> extractRecipient(messageContext, value))
                .collect(Collectors.joining(COMMA));
        }

        return StringUtils.EMPTY;
    }

    private List<Reference> getRecipientsWithoutFirstPractitioner(ReferralRequest referralRequest) {
        boolean firstPractitionerFound = false;
        List<Reference> recipients = new ArrayList<>(referralRequest.getRecipient().size());
        for (Reference reference : referralRequest.getRecipient()) {
            if (!firstPractitionerFound && isReferenceToPractitioner(reference)) {
                firstPractitionerFound = true;
            } else {
                recipients.add(reference);
            }
        }
        return recipients;
    }

    private boolean isReferenceToPractitioner(Reference reference) {
        return reference.getReference().startsWith(ResourceType.Practitioner.name());
    }
```

mattd-kainos (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public static String extractRecipient(MessageContext messageContext, Reference reference) {
+        IIdType referenceId = reference.getReferenceElement();
+        if (referenceId.getResourceType().equals(ResourceType.Practitioner.name())) {
+            return messageContext.getInputBundleHolder()
+                .getResource(referenceId)
+                .map(resource -> (Practitioner) resource)
+                .map(value -> RECIPIENT_PRACTITIONER + extractHumanName(value.getNameFirstRep()))
+                .orElseThrow(() -> new EhrMapperException("Could not resolve Practitioner Reference"));
```
ElliotB-96, line 55: Nice use of exceptions.

ElliotB-96 (approved):

PR #104, opened 2021-02-22
Feature/niad 1019 translate care connect quantity value

mmackiewicz1:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    private static String processCodeableConcept(CodeableConcept value) {
+        if (value.hasCoding() && !value.getCoding().isEmpty()) {
+            Coding coding = value.getCoding().get(0);
```
ElliotB-96, line 98: would it make much difference to use getCodingFirstRep() here?

ElliotB-96 (approved):
Looks good, very well tested tested and all tests are passing in Jenkins, only comment I had I attached.

ElliotB-96 (approved):
Looks well, thank you.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationValueQuantityMapper.java
```diff
@@ -0,0 +1,121 @@
...
+    private ObservationValueQuantityMapper() {
+    }
+
+    public static String processQuantity(Quantity valueQuantity) {
+        String result = StringUtils.EMPTY;
+        if (isUncertaintyCodePresent(valueQuantity)) {
+            result += UNCERTAINTY_CODE;
+        }
+
+        if (!valueQuantity.hasComparator()) {
```
mattd-kainos, line 48: (optional) possible to avoid the negation here

mattd-kainos: I can see how the negation makes it easier to follow along with the written spec though ;)

service/src/main/resources/templates/ehr_observation_statement_effective_time_template.mustache
```diff
@@ -6,11 +6,14 @@
             {{{effectiveTime}}}
         </effectiveTime>
         <availabilityTime value="{{issued}}"/>
+        {{{value}}}
         <pertinentInformation typeCode="PERT">
             <sequenceNumber value="+1" />
             <pertinentAnnotation classCode="OBS" moodCode="EVN">
+                {{{valueQuantity}}}
```
mattd-kainos, line 13: Is this `{{{valueQuantity}}}` used? It doesn't looks like it belongs. I think everything is prepended to the comment

service/src/main/resources/templates/ehr_observation_statement_effective_time_template.mustache
```diff
@@ -6,11 +6,14 @@
             {{{effectiveTime}}}
         </effectiveTime>
         <availabilityTime value="{{issued}}"/>
+        {{{value}}}
```
mattd-kainos, line 9: If I'm reading the code correctly this template is only used by ObservationStatementMapper. ObservationStatementMapper maps unstructured FHIR Observation to HL7 ObservationStatement.

Maybe rename this file to `unstructured_observation_statement_template.mustache`?

service/src/main/resources/templates/ehr_observation_statement_template.mustache
```diff
@@ -6,11 +6,13 @@
             {{#effectiveTime}}<center value="{{effectiveTime}}"/>{{/effectiveTime}}{{^effectiveTime}}<center nullFlavor="UNK"/>{{/effectiveTime}}
         </effectiveTime>
         <availabilityTime value="{{availabilityTime}}" />
+        {{{value}}}
```
mattd-kainos, line 9: If I'm reading the code right this is only used by ImmunizationObservationStatementMapper. I don't think Immunization has any structured value or reference range mappings.

You probably want to revert these template changes.

It would also be nice if you could rename this file to `immunization_observation_statement_template.mustache` to avoid any confusion in the future.

service/src/test/resources/ehr/mapper/observation/pertinent/example-observation-resource-with-pert-info-1.json
```diff
@@ -0,0 +1,59 @@
+{
```
mattd-kainos, line 1: Something to consider for the JSON fixtures for these "sub-mapper". You can probably get away with using minimal resources here e.g. removing all fields except for referenceRange or value[x] as the case may be. It would make these files much smaller.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -0,0 +1,191 @@
...
+                denominator.getValue(),
+                denominator.getUnit());
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private static String processTimeType(TimeType value) {
+        if (value.hasValue()) {
+            return String.format(TIME_VALUE_TEMPLATE, value.getValue());
```
mattd-kainos, line 144: Can we create a stub method in DateFormatUtil to convert BaseDateTimeType into String? It is fine to return getValue() for now, but this will ensure that we update to the correct format and time zone in NIAD-1082

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private static String processDateTimeType(DateTimeType value) {
+        if (value.hasValue()) {
+            return String.format(DATE_TIME_VALUE_TEMPLATE, value.getValue());
```
mattd-kainos, line 152: Can we create a stub method in DateFormatUtil to convert BaseDateTimeType into String? It is fine to return getValue() for now, but this will ensure that we update to the correct format and time zone in NIAD-1082

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        return StringUtils.EMPTY;
+    }
+
+    private static String processPeriod(Period value) {
+        if (value.hasStart() && value.hasEnd()) {
+            return String.format(PERIOD_VALUE_TEMPLATE, value.getStart(), value.getEnd());
```
mattd-kainos, line 160: Can we create a stub method in DateFormatUtil to convert BaseDateTimeType into String? It is fine to return getValue() for now, but this will ensure that we update to the correct format and time zone in NIAD-1082

Use this to create strings for getStartElement() and getEndElement()

mattd-kainos (changes requested):
Very nice PR! This very complex mapping is done in a well structured and reusable way.

Just a few small changes suggested
* template fields that possibly don't need to be there
* date formats: a cause of eternal frustration
* (optional) you could probably trim down some of the JSON test fixtures to be much smaller files

service/src/main/resources/templates/immunization_observation_statement_template.mustache
```diff
@@ -12,5 +12,6 @@
         <text>{{pertinentInformation}}</text>
     </pertinentAnnotation>
 </pertinentInformation>{{/pertinentInformation}}
+        {{{referenceRange}}}
```
mattd-kainos, line 15: I think this {{{referenceRange}}} still needs to be removed. It is not part of Immunization

service/src/test/resources/ehr/mapper/observation/quantity/example-observation-resource-with-quantity-1.json
```diff
@@ -0,0 +1,55 @@
...
+            "reference":"Practitioner/1"
+        }
+    ],
+    "valueQuantity":{
+        "value":37.1,
+        "unit":"C",
+        "system":"http://unitsofmeasure.org",
+        "code":"Cel"
+    }
+}
```
mattd-kainos, line 55: For these mapper tests that are only translating a single property of a resource, I'd strongly recommend trimming down the JSON files to just the relevant parts. It will make our test easier to understand and maintain.

```suggestion
{
    "resourceType":"Observation",
    "valueQuantity":{
        "value":37.1,
        "unit":"C",
        "system":"http://unitsofmeasure.org",
        "code":"Cel"
    }
}
```

mattd-kainos (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    public String mapReferenceRangeToPertinentInformation(
+            Observation.ObservationReferenceRangeComponent observationReferenceRangeComponent) {
+        String result = RANGE_PREFIX;
+        if (observationReferenceRangeComponent.hasText()) {
+            result += TEXT_PREFIX + observationReferenceRangeComponent.getText() + StringUtils.SPACE;
+
+            if (observationReferenceRangeComponent.hasLow()) {
+                SimpleQuantity low = observationReferenceRangeComponent.getLow();
```
msolecka, line 73: 73-80 and 83-92 are very similar, they could be made as function, and you could also consider using builder pattern for that kind of logic, in this case PertinentInformationBuilder

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -1,35 +1,116 @@
...
+        TemplateUtils.loadTemplate("unstructured_observation_statement_template.mustache");
+    private static final String REFERENCE_RANGE_UNIT_PREFIX = "Range Units: ";
+
     private final MessageContext messageContext;
+    private final StructuredObservationValueMapper structuredObservationValueMapper;
+    private final PertinentInformationObservationValueMapper pertinentInformationObservationValueMapper;
 
     public String mapObservationToObservationStatement(Observation observation, boolean isNested) {
-        var observationStatementTemplateParameters = ObservationStatementTemplateParameters.builder()
+        String comment = observation.getComment();
```
msolecka, line 40: What do you think about reversing the mapper logic a bit. It would be more intuitive for me, if there code related to comment, range units and more..  were in separated code chuncks; with separated methods buildComment , buildStructuredElement and all other.

msolecka, line 40: What do you think about reversing the mapper logic a bit. It would be more intuitive for me, if there code related to comment, range units and more.. were in separated code chuncks; with separated methods buildComment , buildStructuredElement and all other.

msolecka (changes requested):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
mmackiewicz1: The problem is that we either construct a comment according to value type in the observation resource or the value element. This is a little bit complicated because we also prepend comment data if we have a reference range present. I am not 100% sure if this solution would make the code more readable or more complicated.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
msolecka: I was thinking about that builder pattern: https://refactoring.guru/design-patterns/builder

mattd-kainos (approved):

msolecka (approved):

PR #125, opened 2021-03-10
NIAD-1113: Map Encounter type to Ehr Composition code

z-uddin:
* Updated EncounterMapper to now translate type field to Ehr Composition code
* Updated existing unit tests that use EncounterMapper to include code field in expected xml
* Added unit tests for EncounterMapper to verify type field is expected code based on mapping logic
* Added resource file for list of Ehr Composition Name vocabulary codes
* Noticed a couple of template parameter classes were in the wrong package and thus moved to relevant package


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -30,9 +47,49 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
...
+            .findFirst()
+            .orElseThrow(() -> new EhrMapperException("Could not map Encounter type"));
+
+        var coding = type.getCoding()
+            .stream()
+            .findFirst()
+            .orElseThrow(() -> new EhrMapperException("Could not map Encounter type coding"));
+
+        if (isSnomedAndWithinEhrCompositionVocabularyCodes(coding)) {
+            return String.format(EHR_COMPOSITION_SNOMED_CODE, coding.getCode(), coding.getDisplay(), buildCodeText(type));
```
msolecka, line 69: Could you use CodeableConceptCdMapper here?

msolecka (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -30,9 +47,49 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
...
+        if (codeableConcept.hasText()) {
+            return String.format(EHR_COMPOSITION_CODE_TEXT, codeableConcept.getText());
+        } else {
+            return StringUtils.EMPTY;
+        }
+    }
+
+    @SneakyThrows
+    private static String[] getEhrCompositionNameVocabularyCodes() {
+        InputStream inputStream = EncounterMapper.class.getClassLoader().getResourceAsStream("ehr_composition_name_vocabulary_codes.txt");
```
mattd-kainos, line 90: Possible improvement: Put these into an unmodifiable Set (HashSet).
Lookups will be must faster. Currently, for every encounter, a new List is created from the array and worse case  contains has to iterate the entire list.
Also the array, although final, could be modified at runtime.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -18,6 +28,13 @@
     private static final Mustache ENCOUNTER_STATEMENT_TO_EHR_COMPOSITION_TEMPLATE =
         TemplateUtils.loadTemplate("ehr_encounter_to_ehr_composition_template.mustache");
     private static final String COMPLETE_CODE = "COMPLETE";
+    private static final String SNOMED_SYSTEM = "http://snomed.info/sct";
+    private static final String EHR_COMPOSITION_SNOMED_CODE = "<code code=\"%s\" displayName=\"%s\" codeSystem"
```
mattd-kainos, line 32: It would be a bit cleaner to create a single mustache template for this, or incorporate into the template for ehrComposition.

mattd-kainos (changes requested):
Should use the mustache template instead of string substitutions when possible. Recommend a performance improvement.

ElliotB-96 (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -30,9 +48,62 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
...
+            .stream()
+            .filter(this::isSnomedAndWithinEhrCompositionVocabularyCodes)
+            .findFirst();
+    }
+
+    private String buildCode(Encounter encounter) {
+        var coding = extractCoding(encounter);
+        if (coding.isPresent()) {
+            return coding.get().getCode();
+        } else {
```
mmackiewicz1, line 81: This "else" is redundant

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -30,9 +48,62 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
...
+            return OTHER_REPORT_CODE;
+        }
+    }
+
+    private String buildDisplayName(Encounter encounter) {
+        var coding = extractCoding(encounter);
+        if (coding.isPresent()) {
+            return coding.get().getDisplay();
+        } else {
```
mmackiewicz1, line 90: This one too.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -30,9 +48,62 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
...
+            return OTHER_REPORT_DISPLAY;
+        }
+    }
+
+    private String buildOriginalText(Encounter encounter) {
+        if (encounter.getTypeFirstRep().hasText()) {
+            return encounter.getTypeFirstRep().getText();
+        } else {
```
mmackiewicz1, line 98: And this

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -48,48 +49,58 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
...
+            return encounter.getTypeFirstRep().getText();
         }
+        return StringUtils.EMPTY;
     }
 
     @SneakyThrows
-    private static String[] getEhrCompositionNameVocabularyCodes() {
-        InputStream inputStream = EncounterMapper.class.getClassLoader().getResourceAsStream("ehr_composition_name_vocabulary_codes.txt");
-        String ehrCompositionNameCodes = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
+    private static HashSet<String> getEhrCompositionNameVocabularyCodes() {
```
mattd-kainos, line 101: A few things to consider here:
* Use the interface (Map) instead of the implementation (HashMap) wherever possible
* Creating the InputStream as a static variable is a bad idea. Nearly all streams need to be closed when you're finished when them. In modern version of java these are "auto-closable" when used in a try clause. In the past you needed to try{...} catch{ ...} finally { stream.close(); }
* Avoid doing things yourself (e.g. splitting by a line) when the standard libraries / language will do this for you. The code might not work as written if the file happens to use windows style line endings.
* Avoid creating objects you don't need. Here we create an array, then a list, and then a hashset. How about just creating the hashmap and adding the elements to it. The Streams API gives us a nice way to collect items into a set.
* 
```
@SneakyThrows
    private static Set<String> getEhrCompositionNameVocabularyCodes() {
        try (InputStream is = EncounterMapper.class.getResourceAsStream("ehr_composition_name_vocabulary_codes.txt")) {
            BufferedReader reader = new BufferedReader(new InputStreamReader(is));
            return reader.lines().collect(Collectors.toSet());
        }
    }
````

Comment (mattd-kainos) [reply service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java:101]: This could be improved further by filtering blank lines:

```
return reader.lines()
                .filter(String::isBlank)
                .collect(Collectors.toSet());
```

Otherwise there is a risk that a blank or whitespace code would be treated as a valid input.

Comment (mattd-kainos) [reply service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java:101]: Finally, upon inspecting the API documentation for the Collectors::toSet method, there is no promise about what kind of set it returns. We do however have an option to create an unmodifiableSet:

```
return reader.lines()
                .filter(String::isBlank)
                .collect(Collectors.toUnmodifiableSet());
```

mattd-kainos (changes requested):
just a fix around handing of input file / input stream

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
z-uddin: Have implemented these changes. 
Needed to add a suppression on RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE to use try-with-resources as there is a known issue with Spotbugs giving false positives.
https://github.com/spotbugs/spotbugs/issues/1338

ElliotB-96 (approved):

mattd-kainos (approved):

PR #132, opened 2021-03-24
NIAD-1181: create component tests for EMIS test extracts

antonlebedjko:
Created ExtractUATTest class which allows us to test importing EhrExtract XML files into EMIS/TPP systems without sending the message through Spine. It takes the EMIS “TC4.zip” (P.S. 9 json files which have been added to a new folder called uat) ASR records and compares the output to expected XML EhrExtract. At the moment, the “expected value” is the current, actual output of the adaptor and this current output has been used to generate 9 stub XML responses files in the uat folder. This will be updated in further tasks.

P.S. I needed to update all 9 json files which all contained the line "description": "" that was causing the fhir validator to throw the error due to the empty description. I have changed it to "description": "none" for now.

Ticket URL: https://gpitbjss.atlassian.net/browse/NIAD-1181


service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,154 @@
...
+        EhrExtractTemplateParameters ehrExtractTemplateParameters = ehrExtractMapper.mapBundleToEhrFhirExtractParams(
+            getGpcStructuredTaskDefinition,
+            bundle);
+        String output = ehrExtractMapper.mapEhrExtractToXml(ehrExtractTemplateParameters);
+
+        //TODO: compare with the proper XML files
+        assertThat(output).isEqualTo(output);
+    }
+
+    private static Stream<Arguments> testValueFilePaths() {
```
n-r-b, line 139: We can use `List.of()` here, then we don't have to wrap each string in `Arguments.of()`

antonlebedjko: I actually just have updated the structure of EhrExtractClass a bit and uploaded some stub XML files for each json we have. So, now the Arguments.of() takes 2 parameters instead of one. The input and output file. Do you think it can stay as it is in this case?

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,165 @@
...
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class EhrExtractUATTest {
+    private static final String TEST_FILE_DIRECTORY = "/uat/";
+    private static final String INPUT_DIRECTORY = "input/";
+    private static final String OUTPUT_DIRECTORY = "output/";
+    private static final String INPUT_PATH = TEST_FILE_DIRECTORY + INPUT_DIRECTORY;
+    private static final String OUTPUT_PATH = TEST_FILE_DIRECTORY + OUTPUT_DIRECTORY;
+    private static final String INPUT_JSON_WITH_PATHOLOGY_RECORD = "TC4 - 9465701718_Guerra_inv1only_20210212.json";
```
saileshp-kainos, line 46: Could we make `TC4 - ` into a constant so it's not repeated throughout the constants? Just a suggestion though & definitely not required

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,165 @@
...
+        String expectedJsonToXmlContent = ResourceTestFileUtils.getFileContent(OUTPUT_PATH + expectedOutputXml);
+        String inputJsonFileContent = ResourceTestFileUtils.getFileContent(INPUT_PATH + inputJson);
+        Bundle bundle = new FhirParseService().parseResource(inputJsonFileContent, Bundle.class);
+        messageContext.initialize(bundle);
+
+        EhrExtractTemplateParameters ehrExtractTemplateParameters = ehrExtractMapper.mapBundleToEhrFhirExtractParams(
+            getGpcStructuredTaskDefinition,
+            bundle);
+        String outputXml = ehrExtractMapper.mapEhrExtractToXml(ehrExtractTemplateParameters);
+        System.out.println(outputXml);
```
saileshp-kainos, line 143: Do we need this `system.out.println`? If not, could it be removed please?

antonlebedjko: Oh, no, we definitely don't need it

antonlebedjko: dfdeccbb374273ec0cf2d19752a525c9b5d564bf

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
Broom94: I would avoid spaces in filenames

antonlebedjko: Does it look better now? 764984545530ece87996d0738d713f8191c29a69 
Thanks for suggestion!

saileshp-kainos: It looks better, but I agree with @Broom94's comment that we should probably avoid spaces in filenames as they can cause weird issues sometimes

antonlebedjko: I agree as well. Here we go: edad4344d51fbe7b44a5d05dc384b0817d43c587

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
n-r-b: Ah ok, no problem 👍

andrewwa-kainos (approved):
You have made a component test, so I've renamed the PR accordingly. LGTM

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,165 @@
...
+        String inputJsonFileContent = ResourceTestFileUtils.getFileContent(INPUT_PATH + FILES_PREFIX + inputJson);
+        Bundle bundle = new FhirParseService().parseResource(inputJsonFileContent, Bundle.class);
+        messageContext.initialize(bundle);
+
+        EhrExtractTemplateParameters ehrExtractTemplateParameters = ehrExtractMapper.mapBundleToEhrFhirExtractParams(
+            getGpcStructuredTaskDefinition,
+            bundle);
+        String outputXml = ehrExtractMapper.mapEhrExtractToXml(ehrExtractTemplateParameters);
+
+        assertThat(outputXml).isEqualToIgnoringWhitespace(expectedJsonToXmlContent);
```
z-uddin, line 145: I think we should not be using isEqualToIgnoringWhitespace 
(as per NIAD-1137)

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final String TEST_TO_ODS_CODE = "test-to-ods-code";
+    private static final String TEST_DATE_TIME = "2020-01-01T01:01:01.01Z";
+
+    private static GetGpcStructuredTaskDefinition getGpcStructuredTaskDefinition;
+
+    @Mock
+    private RandomIdGeneratorService randomIdGeneratorService;
+    @Mock
+    private TimestampService timestampService;
+    @Mock
```
z-uddin, line 87: I don't think we need the Mock annotation for codeableConceptCdMapper

antonlebedjko: If I remove the @mock above the codeableConceptCdMapper declaration, one test will not pass due to the NullPointerException, so seems that it might be needed. What do you think?

antonlebedjko: As has been suggested by you during our call. Here is a change: fc659f6ece666e8a34e241e4b2139509acbf6c6e
Thanks!

andrewwa-kainos (approved):

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
antonlebedjko: Here is a fix cdcaf2f8f3a798de96365af1141d7059f93ade85.
Thanks to @Broom94 for the help! @z-uddin if you are wondering how it has been done because we tried to do it yesterday with you and just copying the output from the terminal didn't work. The way was to remove the XML file, create a new one, copy the output from the terminal in a VSCode (not in IntelliJ) and add the empty line at the end. Only following exactly this pattern has worked. I've tried to it with already existing files and it didn't work. Very weird issue, not really sure why it works in the way it works but now it has been fixed. Thank you!

andrewwa-kainos (approved):

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final String TEST_DATE_TIME = "2020-01-01T01:01:01.01Z";
+
+    private static GetGpcStructuredTaskDefinition getGpcStructuredTaskDefinition;
+
+    @Mock
+    private RandomIdGeneratorService randomIdGeneratorService;
+    @Mock
+    private TimestampService timestampService;
+
+    private CodeableConceptCdMapper codeableConceptCdMapper;
```
n-r-b, line 88: This can be converted to a local variable within `setup()` as it's only used by that method

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @AfterEach
+    public void tearDown() {
+        messageContext.resetMessageContext();
+    }
+
+    @ParameterizedTest
+    @MethodSource("testValueFilePaths")
+    public void When_MappingProperJsonRequestBody_Expect_ProperXmlOutput(String inputJson, String expectedOutputXml) throws IOException {
```
n-r-b, line 135:
```suggestion
    public void When_MappingValidJsonRequestBody_Expect_ValidXmlOutput(String inputJson, String expectedOutputXml) throws IOException {
```

service/src/test/java/uk/nhs/adaptors/gp2gp/uat/EhrExtractUATTest.java
antonlebedjko: Sure! Fixed in 3352717c6bbf17e1c544eb679c56de775e85ae40. Thanks!

n-r-b (dismissed):

andrewwa-kainos (dismissed):

ElliotB-96 (dismissed):
Good job

msolecka (changes requested):
The expected parameter should be wrapped with the <RCMR_IN030000UK06> that means you will need to test the code one step back, 

in GetGpcStructuredTaskExecutor
there is ehr extract created first then it is moved to the final hl7 structure in lines:

```         
String ehrExtractContent = ehrExtractMapper.mapEhrExtractToXml(ehrExtractTemplateParameters);

hl7TranslatedResponse = outputMessageWrapperMapper.map(
      structuredTaskDefinition,
      ehrExtractContent);
``` 

unfortunately they both are done separately to test them together you will need to extract them both to new e.g. Hl7TranslationService

antonlebedjko:
> The expected parameter should be wrapped with the <RCMR_IN030000UK06> that means you will need to test the code one step back,
> 
> in GetGpcStructuredTaskExecutor
> there is ehr extract created first then it is moved to the final hl7 structure in lines:
> 
> ```
> String ehrExtractContent = ehrExtractMapper.mapEhrExtractToXml(ehrExtractTemplateParameters);
> 
> hl7TranslatedResponse = outputMessageWrapperMapper.map(
>       structuredTaskDefinition,
>       ehrExtractContent);
> ```
> 
> unfortunately they both are done separately to test them together you will need to extract them both to new e.g. Hl7TranslationService

@msolecka I did a change here 9f31c29900f213a35091e58e3695aff164ae51c2. Let me know if that way is good enough or not, please. Well spotted, thanks!

msolecka (approved):

andrewwa-kainos (approved):

msolecka:
@antonlebedjko I was thinking about refactoring both mappings by moving them into one new class and testing that one in case there would appear some new mapping between them in the future. But we know what mapping are going to be expanded and nothing like this will happen so seems this solution is good enough. Thank you for the update!

PR #129, opened 2021-03-18
NIAD-1153: gp2gp adaptor uses gpcc adaptor

mmackiewicz1:

PR #135, opened 2021-03-26
NIAD-1125: SDS API requests must use X-Correlation-Id

z-uddin:
* Added X-Correlation-Id header to SdsRequestBuilder
* Added X-Correlation-Id header to Wiremock component tests for SdsClient
* Added component tests that expect exception when X-Correlation-Id header is not a UUID or is not present

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/sds/builder/SdsRequestBuilder.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/sds/builder/SdsRequestBuilder.java
@@ -59,7 +59,8 @@ public class SdsRequestBuilder {
         return buildRequest(task.getFromOdsCode(), RETRIEVE_DOCUMENT_INTERACTION, task.getConversationId());
     }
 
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/sds/builder/SdsRequestBuilder.java
```diff
@@ -37,28 +37,30 @@
         "urn:nhs:names:services:gpconnect:documents:fhir:rest:read:binary-1";
 
     private static final String API_KEY_HEADER = "apikey";
+    private static final String X_CORRELATION_ID_HEADER = "X-Correlation-Id";
```
mmackiewicz1, line 40: This class is being removed in this PR: https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/129. I think the changes should be only done in a consumer.

z-uddin:
Closing this PR as classes SdsClientComponentTest and SdsRequestBuilder will be removed as a part of NIAD-1153 PR #129

PR #139, opened 2021-04-01
NIAD-1024: generate agent directory

ElliotB-96:
* Added AgentDirectoryMapper calls to EhrExtractMapper
* Updated relevant tests


andrewwa-kainos:
Just noticed in the `AgentDirectoryMapper` that there's a set of strings field `mappedOrganizationsAndPractitioner` which is going to grow and grow in size over the application's lifetime. This should probably be moved to be an instance variable passed between the relevant methods, so that it's GCed after the mapper is done.

If you don't want to pass around extra variables, you could use an inner class to do the actual mapping, and that could hold state as it itself would be GCed.

andrewwa-kainos (approved):

msolecka (approved):

PR #143, opened 2021-04-07
NIAD-1284 Default paths and wiremock conform to GP Connect specifications

mattd-kainos:


mattd-kainos:
Not needed - changes made in other PRs

PR #151, opened 2021-04-09
NIAD-1096: ACs 4, 5, 6 related to mapping from ReferallRequest to RequestStatement

andrewwa-kainos:
AC4 - add Participant element to RequestStatement

GIVEN ReferralRequest is translated to RequestStatement
AND ReferralRequest.requester is translated to Agent
THEN Practitioner element is added to RequestStatement

AC5 - add responsibleParty/agentRef element to RequestStatement

GIVEN ReferralRequest is translated to RequestStatement
AND ReferralRequest.recipient[0] is translated to Agent
THEN responsibleParty/AgentRef element is added to RequestStatement
AND AgentRef is referencing Agent

AC6 - add text element to RequestStatement

GIVEN ReferralRequest is translated to RequestStatement
THEN add Practitioner.name + “ “ + Practitioner.family text to RequestStatement.text for every Practitioner resource referenced by ReferralRequest.recipient[1..n]


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -58,21 +62,35 @@
...
 
         if (!referralRequest.hasReasonCode()) {
-            requestStatementTemplateParameters.setDefaultReasonCode(DEFAULT_REASON_CODE_XML);
+            requestStatementTemplateParameters.defaultReasonCode(DEFAULT_REASON_CODE_XML);
         }
 
-        return TemplateUtils.fillTemplate(REQUEST_STATEMENT_TEMPLATE, requestStatementTemplateParameters);
+        if (referralRequest.hasRequester() && referralRequest.getRequester().hasAgent()
+                && referralRequest.getRequester().getAgent().hasReference()) {
+            final String agentRef = idMapper.get(referralRequest.getRequester().getAgent());
```
n-r-b, line 82: How do we know if the `requester.agent` is a practitioner reference?

andrewwa-kainos: For my reference:

> if requester.agent is Device reference then generate a string of the form "Requester Device: { human readable Device.code text i.e. term text } " and prepend to RequestStatement/text
> 
> if requester.agent is Organisation reference then generate a string of the form "Requester Org: { Referenced Organisation.name } " and prepend to RequestStatement/text
> 
> if requester.agent is Patient reference then generate a string of the form "Requester: Patient " and prepend to RequestStatement/text
> 
> if requester.agent is RelatedPerson reference then generate a string of the form "Requester: Relation {RelatedPerson.name} " and prepend to RequestStatement/text
> 
> If requester.agent is Practitioner Reference then build a Participant referencing an agentPerson constructed from the requester.agent reference and the requester.onBehalfOf Organization if populated. If there is no requester.onBehalfOfOrganization then fallback to the Organisation on PracttitionerRole or Patient.managingOrganisation
> 
> <Participant typeCode="PRF" contextControlCode="OP"><agentRef classCode="AGNT"><id root="Uuid for participant"/></agentRef></Participant>

andrewwa-kainos: I have all of the above prepared except for tests and the final point regarding `onBehalfOf`. That one will need some changes made to the agent directory, tbd in another story ∴ blocking this.

msolecka: I also understand it the same way as Natalie, I think we should check here if the agent is the reference to Practitioner and only then generate participant, because only for requester in the type of Practitioner the agentPerson is generated in AgentDirectory.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -213,6 +232,26 @@ private String buildTextDescription(ReferralRequest referralRequest) {
...
+                .collect(Collectors.toList())
+            : Collections.emptyList();
+    }
+
+    private String getRecipientName(Reference reference) {
+        final IIdType iidType = reference.getReferenceElement();
+        return messageContext.getInputBundleHolder()
+            .getResource(iidType)
+            .map(Practitioner.class::cast)
+            .map(practitioner -> practitioner.getNameFirstRep().getNameAsSingleString())
```
msolecka, line 251: I think this method will map to string also all prefixes and all suffixes. The story says only about name and family name.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -213,6 +232,26 @@ private String buildTextDescription(ReferralRequest referralRequest) {
         return text.orElse(StringUtils.EMPTY);
     }
 
+    private List<String> buildPractitionersDescription(ReferralRequest referralRequest) {
+        return referralRequest.hasRecipient()
+            ? referralRequest.getRecipient().stream()
```
msolecka, line 237: Before each recipient there should be a prefix _**Recipient Practitioner:**_
```
 Recipient Practitioner: { Practitioner.name + " " + Practitioner.family}
```

for example: 
Recipient Practitioner: Arthur George

I am not sure if there would be more how should they be handled?

```
Recipient Practitioner: Arthur George Recipient Practitioner: John Abc
```
or 

```
Recipient Practitioner: Arthur George, John Abc
```
? 
Could you please also add test cases to cover multiple case?

andrewwa-kainos: Just noticed there seems to be some duplication here with NIAD-916, where various data is being output in sequence without braces. E.g.
`UBRN: bb2378b6-dde2-11e9-9d36-2a2ae2dbcce4 Priority: Routine Specialty: unknown Recipient Org: Organisation name, Recipient Healthcare Service: Consulting psychologists and/or psychology services, Recipient Practitioner: Arthur George Referred for investigation into low blood pressure`

andrewwa-kainos: I've updated this all to follow the pattern requested in the newer ticket

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
```diff
@@ -50,20 +47,20 @@ public static String extractRecipient(MessageContext messageContext, Reference r
         if (referenceId.getResourceType().equals(ResourceType.Practitioner.name())) {
             return messageContext.getInputBundleHolder()
                 .getResource(referenceId)
-                .map(resource -> (Practitioner) resource)
-                .map(value -> RECIPIENT_PRACTITIONER + extractHumanName(value.getNameFirstRep()))
+                .map(Practitioner.class::cast)
+                .map(value -> "Recipient Practitioner: { " + extractHumanName(value.getNameFirstRep()) + " }")
```
mattd-kainos, line 51: Omit the {}'s. These are meant to indicate a substitution in the spec pseudocode.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
```diff
@@ -50,20 +47,20 @@ public static String extractRecipient(MessageContext messageContext, Reference r
...
                 .orElseThrow(() -> new EhrMapperException("Could not resolve Practitioner Reference"));
         } else if (referenceId.getResourceType().equals(ResourceType.HealthcareService.name())) {
             return messageContext.getInputBundleHolder()
                 .getResource(referenceId)
-                .map(resource -> (HealthcareService) resource)
-                .map(value -> RECIPIENT_HEALTH_CARE_SERVICE + value.getName())
+                .map(HealthcareService.class::cast)
+                .map(value -> "Recipient Healthcare Service: { " + value.getName() + " }")
```
mattd-kainos, line 57: as above

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementExtractor.java
```diff
@@ -50,20 +47,20 @@ public static String extractRecipient(MessageContext messageContext, Reference r
...
                 .orElseThrow(() -> new EhrMapperException("Could not resolve HealthcareService Reference"));
         } else if (referenceId.getResourceType().equals(ResourceType.Organization.name())) {
             return messageContext.getInputBundleHolder()
                 .getResource(referenceId)
-                .map(resource -> (Organization) resource)
-                .map(value -> RECIPIENT_ORG + value.getName())
+                .map(Organization.class::cast)
+                .map(value -> "Recipient Org: { " + value.getName() + " }")
```
mattd-kainos, line 63: as above

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,38 +41,101 @@
...
-    private static final String REQUESTER_DEVICE = "Requester Device: ";
-    private static final String REQUESTER_ORG = "Requester Org: ";
     private static final String REQUESTER_PATIENT = "Requester: Patient";
-    private static final String REQUESTER_RELATION = "Requester: Relation ";
-    private static final String SPECIALTY = "Specialty: ";
-    private static final String REASON_CODE = "Reason Codes: ";
     private static final String DEFAULT_REASON_CODE_XML = "<code code=\"3457005\" displayName=\"Patient referral\" codeSystem=\"2.16.840.1"
         + ".113883.2.1.3.2.4.15\"/>";
-    private static final String NOTE = "Annotation: %s @ %s %s";
+    private static final String NOTE = "Annotation: { %s @ %s %s }";
```
mattd-kainos, line 48: as above, {}'s should be omitted

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/parameters/RequestStatementTemplateParameters.java
```diff
@@ -11,7 +17,19 @@
...
     private String availabilityTime;
-    private String description;
     private String defaultReasonCode;
     private String code;
+    private String participant;
+    private String responsibleParty;
+    @Singular
+    private List<String> texts;
+
+    public Function<Object, String> compileText() {
```
mattd-kainos, line 27: Why is this this reversed? I don't think this logic should be in our parameters object. All other parameters objects are logic-free.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: This should now be covered

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: We're now using `RequestStatementExtractor::extractHumanName`

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,38 +41,101 @@
...
+                .map(Device.class::cast)
+                .filter(Device::hasType)
+                .map(Device::getType)
+                .flatMap(CodeableConceptMappingUtils::extractTextOrCoding)
+                .map("Requester Device: "::concat)
+                .ifPresentOrElse(requestStatementTemplateParameters::text, () -> {
+                    throw new EhrMapperException("Could not resolve Device Reference");
+                });
+
+        } else if (isReferenceToType(agent, ResourceType.Organization)) {
```
n-r-b, line 112: Would we ever reach this? Line 95 has the same check

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/parameters/RequestStatementTemplateParameters.java
andrewwa-kainos: This seemed like the simplest way to build it - add to the list and then reverse it to simulate prepending. Fair shout re: logic. Is the only requirement that `ReferralRequest.description` be last, and all other content may be unordered before it?

mattd-kainos: > the only requirement that ReferralRequest.description be last, and all other content may be unordered before it

that is correct

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: Well spotted! I think actually those first two conditions are separate from the rest and the chain could be broken there. Then just need to make sure the exception is thrown in the right conditions

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,182 +41,244 @@
...
-                return StringUtils.EMPTY;
+        private String buildIdentifierDescription() {
+            return Optional.of(referralRequest).stream()
+                .filter(ReferralRequest::hasIdentifier)
+                .map(ReferralRequest::getIdentifier)
+                .flatMap(List::stream)
+                .filter(Objects::nonNull)
+                .filter(val -> UBRN_SYSTEM_URL.equals(val.getSystem()))
+                .map(Identifier::getValue)
+                .findAny()
```
n-r-b, line 196: We were finding the first element previously, does `.findAny()` change anything?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,182 +41,244 @@
...
+
+            } else if (isReferenceToType(agent, ResourceType.RelatedPerson)) {
+                messageContext.getInputBundleHolder().getResource(agent.getReferenceElement())
+                    .filter(RelatedPerson.class::isInstance)
+                    .map(RelatedPerson.class::cast)
+                    .filter(RelatedPerson::hasName)
+                    .map(RelatedPerson::getNameFirstRep)
+                    .map(RequestStatementExtractor::extractHumanName)
+                    .map(REQUESTER_RELATION::concat)
+                    .ifPresent(this::prependText);
```
n-r-b, line 161: I believe this exception was there previously:
```suggestion
                    .ifPresentOrElse(this::prependText, () -> {
                        throw new EhrMapperException("Could not resolve RelatedPerson Reference");
                    });
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,182 +41,244 @@
...
+                    });
+
+            } else if (isReferenceToType(agent, ResourceType.Organization)) {
+                messageContext.getInputBundleHolder().getResource(agent.getReferenceElement())
+                    .filter(Organization.class::isInstance)
+                    .map(Organization.class::cast)
+                    .filter(Organization::hasName)
+                    .map(Organization::getName)
+                    .map(REQUESTER_ORG::concat)
+                    .ifPresent(this::prependText);
```
n-r-b, line 145:
```suggestion
                    .ifPresentOrElse(this::prependText, () -> {
                        throw new EhrMapperException("Could not resolve Organization Reference");
                    });
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: In most cases they're equivalent, pragmatically; for this code, we're looking at the values of identifiers with the UBRN system which I'd expect there only to be one of in good data. In dodgy data with multiples... they _should_ have the same value, right? These things are meant to identify the thing they represent after all. If so, findAny or findFirst are equivalent, and if not then... GIGO?

service/src/test/resources/ehr/mapper/referral/expected-output-request-statement-9.xml
```diff
@@ -6,6 +6,6 @@
         </effectiveTime>
         <availabilityTime value="20100119"/>
         <code code="3457005" displayName="Patient referral" codeSystem="2.16.840.1.113883.2.1.3.2.4.15"/>
-        <text>UBRN: bb2378b6-dde2-11e9-9d36-2a2ae2dbcce4 Priority: Routine Requester: Patient Specialty: unknown Referred for investigation into low blood pressure</text>
+        <text>UBRN: bb2378b6-dde2-11e9-9d36-2a2ae2dbcce4 Priority: Routine Specialty: unknown Referred for investigation into low blood pressure</text>
```
msolecka, line 9: looking at data and I think that case 9 should have` Requester: Patient` still, it has in ReferralRequest:

```
  "requester": {
        "agent": {
            "reference": "Patient/4C903809-CADA-442E-8FA5-655E5EFEB1F5"
        }
    }```

File: service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java [hunk 14]
```diff
@@ -41,182 +42,238 @@
...
+                templateParameters.defaultReasonCode(DEFAULT_REASON_CODE_XML);
+            }
+
+            if (referralRequest.hasRequester()
+                    && referralRequest.getRequester().hasAgent()
+                    && referralRequest.getRequester().getAgent().hasReference()) {
+                var requester = referralRequest.getRequester();
+                Reference agentRef = requester.getAgent();
+                var onBehalfOf = requester.hasOnBehalfOf() ? requester.getOnBehalfOf() : null;
+                processAgent(agentRef, onBehalfOf);
```
msolecka, line 97: what do you think about changing the method processAgent into 2 : getAgent and getText ?  I think process agent method has too much responsibilities at the moment.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,182 +42,238 @@
...
+            }
+
+            if (referralRequest.hasRecipient()) {
+                final Reference recipient = referralRequest.getRecipientFirstRep();
+                final var responsibleParty = idMapper.get(recipient);
+                templateParameters.responsibleParty(responsibleParty);
+            }
+
+            templateParameters.text(text);
```
msolecka, line 106: For me it is not easy to see where all parts of the text are prepared, the methods have "unexpected side effect"  - building the text is hidden inside different methods.  I would try to build text without state, prepare one method that will take a resource, apply logic and return string value that could be placed in the template. I would stay with the way it was implemented in buildDescription and extend with additional new logic.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,182 +42,238 @@
...
-            .code(buildCode(referralRequest))
-            .build();
-
-        if (!referralRequest.hasReasonCode()) {
-            requestStatementTemplateParameters.setDefaultReasonCode(DEFAULT_REASON_CODE_XML);
+        return new InnerMapper(referralRequest, isNested).map();
+    }
+
+    @RequiredArgsConstructor
+    private class InnerMapper {
```
msolecka, line 70: What is the purpose of the InnerMapper ?

andrewwa-kainos: A private inner class is a handy way of letting a singleton unsynchronized class do its work with thread-safe state. Helps to simplify code (no need to pass around parameters as much as they can be taken from instance fields instead). I introduced it because of the complexity around creating the text, but even if I change that process as per your other comment I might still prefer to keep the inner class. What do you think?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: Agreed. This has been refactored - still pretty involved, but hopefully clearer than before.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: Agreed. The main difficulty was in maintaining that exception thrown for an unexpected agent type; please let me know what you think

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
msolecka: sure, I was not familiar with that approach, we can keep it.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
n-r-b: Ok, fair enough 👍

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -41,182 +44,265 @@
...
+            return StringUtils.EMPTY;
+        }
 
-    private String buildReasonCodeDescription(ReferralRequest referralRequest) {
-        if (referralRequest.hasReasonCode() && (referralRequest.getReasonCode().size() > 1)) {
-            return REASON_CODE + extractReasonCode(referralRequest);
+        private String buildPatientDescription(Reference agent) {
+            if (isReferenceToType(agent, ResourceType.Patient)) {
+                return messageContext.getInputBundleHolder().getResource(agent.getReferenceElement())
+                    .filter(Patient.class::isInstance)
```
msolecka, line 239: I don't think it is necessary to check if the Patient resource is in the bundle. Checking the reference type seems to be enough, because we do not need any information from Patient resource to extend text.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
msolecka: I prefer the upgraded option now, thank you for changing it.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
andrewwa-kainos: Sure, I'll remove all the filters doing that

service/src/test/resources/ehr/mapper/referral/expected-output-request-statement-9.xml
andrewwa-kainos: It doesn't because the input fhir-bundle.json doesn't have this. I'll add it to the bundle

msolecka (approved):

mattd-kainos (approved):

PR #152, opened 2021-04-09
NIAD-1096: Add AUT and PRF practitioners to ObservationStatement when mapping allergy intolerance

andrewwa-kainos:
https://gpitbjss.atlassian.net/browse/NIAD-1096

AC7 - add Participant element to ObservationStatement

GIVEN AllergyIntolerance is translated to ObservationStatement
THEN Practitioner element of type ”AUT” is added to ObservationStatement for AllergyIntolerance.recorder
AND Practitioner element of type ”PRF” is added to ObservationStatement for AllergyIntolerance.asserter
AND Practitioner is referencing Agent


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -26,6 +27,14 @@ public String getOrNew(Reference reference) {
         return mappedId;
     }
 
+    public String get(Reference reference) throws EhrMapperException {
+        String mappedId = ids.get(reference.getReference());
```
msolecka, line 31: I don't think you should add the method that expects id mapped to be already mapped when mapping AllergyIntolerance asserter and recorder. What I mean, I think we cannot be sure that the resources will be in order ( that referencing Practitioner will be mapped before AllergyIntolerance ), but we were assured that the referencing object will exist in the bundle.  So I suggest to always use method getOrNew that will generate and store the id for practitioner whenever it was reference in AllergyIntolerance or Practitioner id mapped first.

andrewwa-kainos: In general for these stories, these mappers are indeed assuming that the agent directory has been processed already. In that case, references to agents that haven't been mapped into the directory (and added to this `IdMapper`) _should_ throw an error.

@mattd-kainos can you confirm or clarify please?

msolecka: ah, you're right, agent Directory is done first. I thought the order is different. Sorry.

msolecka (approved):

msolecka (approved):

mattd-kainos (approved):

PR #153, opened 2021-04-09
NIAD-1096: Add participant as practitioner when mapping condition to link set

andrewwa-kainos:
https://gpitbjss.atlassian.net/browse/NIAD-1096

AC8 - add Participant element to LinkSet

GIVEN Condition is translated to LinkSet
THEN Practitioner element is added to LinkSet
AND Practitioner is referencing Agent


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapper.java
```diff
@@ -72,6 +74,13 @@ public String mapConditionToLinkSet(Condition condition, boolean isNested) {
 
         builder.code(buildCode(condition));
 
+        if (condition.hasAsserter()) {
+            var asserterReference = condition.getAsserter();
```
msolecka, line 78: it would be good to check here is the asserter is referencing Practitioner not other resource.

msolecka: Could you please also add test to not map the asserter if it is not practitioner ?

service/src/test/resources/ehr/mapper/condition/expected_output_linkset_1.xml
```diff
@@ -8,22 +8,22 @@
         <availabilityTime value="20200907101202" />
         <conditionNamed typeCode="NAME" inversionInd="true">
             <namedStatementRef classCode="OBS" moodCode="EVN">
-                <id root="7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB"/>
+                <id root="Condition/7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB"/>
```
msolecka, line 11: I know it is just the matter of mock, but it will be misleading for somebody reading that test, that the xml contains fhir reference. I think it is good idea to solve mocking mapperId the way you proposed. Maybe we could just change the id to not look like fhir reference, for example moving resource name at the end? 
` 7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB-Condition `

msolecka (changes requested):

service/src/test/resources/ehr/mapper/condition/expected_output_linkset_1.xml
msolecka: yes, it's very good idea to mock it your way, we really needed that, now we can see if the ids are correctly mapped!

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapper.java
mattd-kainos: only practitioner is profiled, so this would be unlikely to occur

service/src/test/resources/ehr/mapper/condition/expected_output_linkset_1.xml
mattd-kainos: 7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB-Condition is actually kind of how EMIS generate their FHIR identifiers.

This is not, however, valid HL7 which requires real UUIDs. Maybe ok if we make it obvious we are using intentionally invalid identifiers? Related tech debt https://gpitbjss.atlassian.net/browse/NIAD-1293

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapper.java
```diff
@@ -72,6 +74,15 @@ public String mapConditionToLinkSet(Condition condition, boolean isNested) {
 
         builder.code(buildCode(condition));
 
+        if (condition.hasAsserter()) {
```
n-r-b, line 77: I don't see any tests for a condition without an asserter? Otherwise, if this is mandatory, maybe we should throw an exception when there's no asserter?

service/src/test/resources/ehr/mapper/condition/expected_output_linkset_1.xml
msolecka: if we use the same id for each resource we won't see if there would be an error with for example assigning wrong reference or assigning the same id twice, especially in a huge bundle.

If that is emis identifier we could enrich the name with "test" or "mapper" or "new" suffix, for example 7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB-Condition-new-id ?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapper.java
z-uddin: I also agree, believe asserter is a mandatory field (from the GP connect API https://gpc-structured-1-5-0.netlify.app/accessrecord_structured_problems.html) thus should throw an exception when asserter is not present

mattd-kainos: Similarly, only Practitioner is profiled for the reference. Since we check the reference type in line 80 we should also throw and exception if its not a practitioner instead of silently skipping the field.

service/src/test/resources/ehr/mapper/condition/expected_output_linkset_1.xml
andrewwa-kainos: Any decision on this, @msolecka @mattd-kainos ?

msolecka: we can go with 7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB-Condition-new-id

mattd-kainos: I agree here that this convention of using "ResourceType/UUID" will confuse in the future because it looks like a FHIR reference. A suffix as suggested by Gosia maintains this traceability through the test fixtures without making the value appear to be something it is not.

We begin to get a bit verbose now but the information we're trying to convey is essentially `7E277DF1-6F1C-47CD-84F7-E9B7BF4105DB-was-generated-for-Condition`.

Note this still is not a valid HL7 instance identifier (II) but that is ok since these are mocks.

andrewwa-kainos: Substituted `root="([^"/]+)/([^"]+)"` with `root="$2-$1-new-ID"` in all expected_output_linkset_*.xml files

msolecka (approved):

n-r-b (approved):

PR #159, opened 2021-04-15
NIAD-1096: Add practitioner/practitionerRole/organization recorder when mapping medication request to medication statement

andrewwa-kainos:
AC11 - add recorder element to MedicationStatement

GIVEN MedicationRequest is translated to MedicationStatement
WHEN MedicationRequest.Recorder has practitioner/**patient**~~/practitonerRole/Organisation~~ reference
THEN generate Participant at MedicationStatement level corresponding to that practitioner reference with typeCode “AUT”


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
```diff
@@ -243,4 +255,20 @@ private String buildPriorPrescription(MedicationRequest medicationRequest) {
...
+        Predicate<String> isRelevant = isPractitioner.or(isPractitionerRole).or(isOrganization);
+
+        if (medicationRequest.hasRecorder() && medicationRequest.getRecorder().hasReference()) {
+            final var reference = medicationRequest.getRecorder();
+            if (isRelevant.test(reference.getReference())) {
+                return participantMapper.mapToParticipant(
+                    messageContext.getIdMapper().get(reference), ParticipantType.AUTHOR);
+            }
+        }
+        return StringUtils.EMPTY;
```
mattd-kainos, line 272: Only Practitioner and and Patient are profiled, so recorder cannot reference a PractitionerRole or organisation here.

So anything other than Practitioner or Patient we should raise an error.

For patient: 

> if   recorder is Patient  then prepend   string "Recorded By Patient " to pertinentInformation text

andrewwa-kainos: Is the AC wrong?

> WHEN MedicationRequest.Recorder has practitioner/practitonerRole/Organisation reference

Or the others just need to be dealt with another way?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
```diff
@@ -243,4 +255,20 @@ private String buildPriorPrescription(MedicationRequest medicationRequest) {
         }
         return StringUtils.EMPTY;
     }
+
+    private String buildParticipant(MedicationRequest medicationRequest) {
+        Predicate<String> isPractitioner = s -> s.startsWith(ResourceType.Practitioner.name());
+        Predicate<String> isPractitionerRole = s -> s.startsWith(ResourceType.PractitionerRole.name());
+        Predicate<String> isOrganization = s -> s.startsWith(ResourceType.Organization.name());
```
andrewwa-kainos, line 262: These aren't right, because `"PractitionerRole/UUID".startsWith("Practitioner")`

msolecka: to avoid it, there could be used, reference.getReferenceElement().getResourceType() and then verify if it is equal.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
mattd-kainos: Yes it looks like the AC is wrong in this case. The mapping spec is always the "source of truth" of they don't agree

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
```diff
@@ -243,4 +257,23 @@ private String buildPriorPrescription(MedicationRequest medicationRequest) {
...
     }
+
+    private String buildParticipant(MedicationRequest medicationRequest) {
+        Predicate<Reference> isPractitioner = buildPredicateReferenceIsA(ResourceType.Practitioner);
+        Predicate<Reference> isPatient = buildPredicateReferenceIsA(ResourceType.Patient);
+        Predicate<Reference> isRelevant = isPractitioner.or(isPatient);
+
+        if (medicationRequest.hasRecorder() && medicationRequest.getRecorder().hasReference()) {
+            final var reference = medicationRequest.getRecorder();
+            if (reference.hasReferenceElement() && isRelevant.test(reference)) {
```
z-uddin, line 268: From the earlier discussion, shouldn't we only map to participant when the recorder reference is a Practitioner.
And if the recorder we generate some text prepended to the PertinentInformation?
(from @mattd-kainos 's comments on the updated AC)

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
andrewwa-kainos: Which pertinentInformation does that refer to @mattd-kainos ?

There's:
`MedicationStatement.pertinentInformation.pertinentMedicationDosage`
`MedicationStatement.component.ehrSupplyAuthorise.pertinentInformation.pertinentSupplyAnnotation`
and the spreadsheet also describes this for `note`:
```
<pertinentInformation typeCode="PERT">
  <sequenceNumber value="+1"/>
  <pertinentAnnotation classCode="OBS" moodCode="EVN">
    <text>{note}</text>
  </pertinentAnnotation>
</pertinentInformation>
```
but which is not yet in the relevant mustache template

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
andrewwa-kainos: Yes. Waiting to find out which pertinentInformation block it should be added to

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
mattd-kainos: It seems I was looking at allergy instead of medication when I wrote this. MedicationRequest:G36 is the right place to look

recorder | I | 1..1 | Reference ( CareConnect-GPC-Practitioner-1 ) | Person who entered the request | y | If Reference(Practitioner)   populated generate an agentPerson Agent corresponding to the   Practitioner/PractitionerRole/Organisation and reference as a Participant of   type AUT at MedicationStatement level - not ehrSupplyX level     <Participant typeCode="AUT"   contextControlCode="OP"><agentRef   classCode="AGNT"><id root="Uuid for   participant"/></agentRef></Participant>
-- | -- | -- | -- | -- | -- | --

ElliotB-96 (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
```diff
@@ -243,4 +257,24 @@ private String buildPriorPrescription(MedicationRequest medicationRequest) {
...
+
+    private String buildParticipant(MedicationRequest medicationRequest) {
+        var isPractitioner = buildPredicateReferenceIsA(ResourceType.Practitioner);
+        var isPractitionerRole = buildPredicateReferenceIsA(ResourceType.PractitionerRole);
+        var isOrganization = buildPredicateReferenceIsA(ResourceType.Organization);
+        Predicate<Reference> isRelevant = isPractitioner.or(isPractitionerRole).or(isOrganization);
+
+        if (medicationRequest.hasRecorder() && medicationRequest.getRecorder().hasReference()) {
+            final var reference = medicationRequest.getRecorder();
+            if (reference.hasReferenceElement() && isRelevant.test(reference)) {
```
n-r-b, line 269: Is `reference.hasReferenceElement()` not the same as `medicationRequest.getRecorder().hasReference()`?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
```diff
@@ -243,4 +257,24 @@ private String buildPriorPrescription(MedicationRequest medicationRequest) {
...
+                return participantMapper.mapToParticipant(
+                    messageContext.getIdMapper().get(reference), ParticipantType.AUTHOR);
+            }
+        }
+        return StringUtils.EMPTY;
```
n-r-b, line 274: I think we're missing a test covering a `medicationRequest` without a reference, and returning `StringUtils.EMPTY`

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
mattd-kainos: I'm just guessing, but this may be because HAPI tries to be "helpful" and create empty instances for you if you call getAbc() for abc that doesn't exist.

andrewwa-kainos: You're right @n-r-b, it's redundant. Will remove

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/MedicationStatementMapper.java
andrewwa-kainos: A recorder without a reference would be an empty object in the json. Do we still want to test for it? There is a precedent; search regex `" ?: ?\{\s*}` gives 12 matches, mostly from document reference tests

z-uddin: Just to add, from the GP connect API (https://gpc-structured-1-5-0.netlify.app/accessrecord_structured_development_medicationrequest.html)
recorder is a mandatory field, thus should we throw an error if recorder is not populated / does not have a reference?

z-uddin (approved):

mattd-kainos (approved):

n-r-b (approved):

PR #167, opened 2021-04-20
NIAD-1306: Generic handling of Observation component

z-uddin:
* Now translate component field in Observation resource, prepending text to pertinent information in HL7
* Replaced StringBuilder inserts to appends in ObservationStatementMapper.prepareComment
* Added unit tests in ObservationStatementMapperTest


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -38,6 +42,11 @@
             .put(DateTimeType.class, value -> processDateTimeType((DateTimeType) value))
             .put(Period.class, value -> processPeriod((Period) value))
             .build();
+    private static final Map<Class<? extends IBaseElement>, Function<IBaseElement, String>> COMPONENT_VALUE_FUNCTIONS =
+        ImmutableMap.<Class<? extends IBaseElement>, Function<IBaseElement, String>>builder()
+            .put(Quantity.class, value -> processComponentValueQuantity((Quantity) value))
+            .put(StringType.class, value -> processComponentValueString((StringType) value))
+            .build();
```
andrewwa-kainos, line 49: `Map.of()` is a lot simpler, imo

```suggestion
    private static final Map<Class<? extends IBaseElement>, Function<IBaseElement, String>> COMPONENT_VALUE_FUNCTIONS = Map.of(
            Quantity.class, value -> processComponentValueQuantity((Quantity) value),
            StringType.class, value -> processComponentValueString((StringType) value)
    );
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -200,4 +217,144 @@ private static boolean isDenominatorPresent(Ratio ratio) {
...
+
+                if (userSelectedCoding.isPresent()) {
+                    return String.format(COMPONENT_INTERPRETATION_CODE_TEMPLATE, userSelectedCoding.get().getCode());
+                } else {
+                    return String.format(COMPONENT_INTERPRETATION_CODE_TEMPLATE,
+                        component.getInterpretation().getCodingFirstRep().getCode());
+                }
+            }
+        }
+        return StringUtils.EMPTY;
```
andrewwa-kainos, line 292: The two nested `if`s can be combined, and there's some optimisation we can make to the coding selection, I think:

```suggestion
        if (component.hasInterpretation() && component.getInterpretation().hasCoding()) {
            var userSelectedCoding = component
                .getInterpretation()
                .getCoding()
                .stream()
                .filter(Coding::hasUserSelected)
                .findFirst();
            
            var codingTouse = userSelectedCoding.orElseGet(() -> component.getInterpretation().getCodingFirstRep());
            
            return String.format(COMPONENT_INTERPRETATION_CODE_TEMPLATE, codingTouse.getCode());
        }
        return StringUtils.EMPTY;
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -200,4 +215,139 @@ private static boolean isDenominatorPresent(Ratio ratio) {
...
+
+    private String extractComponentValue(Observation.ObservationComponentComponent component) {
+        if (component.hasValue()) {
+            var value = component.getValue();
+
+            if (COMPONENT_VALUE_FUNCTIONS.containsKey(value.getClass())) {
+                return COMPONENT_VALUE_FUNCTIONS.get(value.getClass()).apply(value);
+            } else {
+                return mapObservationValueToPertinentInformation(value);
+            }
```
andrewwa-kainos, line 267: So this is pretty much `.getOrDefault` but it's a bit awkward because the default doesn't need the key (class)

```suggestion
            return COMPONENT_VALUE_FUNCTIONS.getOrDefault(value.getClass(),
                $ -> this.mapObservationValueToPertinentInformation(value)
            ).apply(value);
```

Your choice

andrewwa-kainos (approved):
One more optional observation, but LGTM

service/src/test/resources/ehr/mapper/observation/expected-output-observation-statement-25.xml
```diff
@@ -0,0 +1,20 @@
...
+        <statusCode code="COMPLETE" />
+        <effectiveTime>
+            <center value="20100714163251"/>
+        </effectiveTime>
+        <availabilityTime value="20100113152950"/>
+        
+        <pertinentInformation typeCode="PERT">
+            <sequenceNumber value="+1" />
+            <pertinentAnnotation classCode="OBS" moodCode="EVN">
+                <text>Component(s): [Code: Allergy to penicillin Code Value: test-code Test display  Interpretation Code: Example without userSelected Interpretation Text: Example interpretation text 1 Range: 20 test-unit 30 test-unit] [Code: BMD - Left Femur String Value: test-string Interpretation Code: Example with userSelected Interpretation Text: Example interpretation text 2] [String Value: Example of a valueString] [Quantity Value: 7.2 g/dl] Example comment</text>
```
msolecka, line 13: I am not sure the Code Value should be in observation pertinentInformation. Also there is double space after.

service/src/test/resources/ehr/mapper/observation/expected-output-observation-statement-22.xml
```diff
@@ -0,0 +1,20 @@
...
+        <statusCode code="COMPLETE" />
+        <effectiveTime>
+            <center value="20100714163251"/>
+        </effectiveTime>
+        <availabilityTime value="20100113152950"/>
+        
+        <pertinentInformation typeCode="PERT">
+            <sequenceNumber value="+1" />
+            <pertinentAnnotation classCode="OBS" moodCode="EVN">
+                <text>Component(s): [Code Value: test-code Example display valueCodeableConcept ] [Range Value: Low 10.1 test-unit High 20.5 test-unit ] [Ratio Value: &lt; 10 test-unit / &gt; 20 test-unit ] [Time Value: 12:30:10 ] [DateTime Value: 2017-03-01 12:50:12 ] [Period Value: Start 2017-03-01 12:50:12 End 2018-06-03 14:12:10 ] Example comment</text>
```
msolecka, line 13: all the [] has additional space before the end.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -96,6 +96,11 @@ public String mapObservationToObservationStatement(Observation observation, bool
     private String prepareComment(Observation observation) {
         StringBuilder commentBuilder = new StringBuilder(observation.hasComment() ? observation.getComment() : StringUtils.EMPTY);
 
+        if (observation.hasComponent()) {
+            commentBuilder.insert(COMMENT_OFFSET,
```
mattd-kainos, line 100: Optional improvement, but boy scout rule :) Instead of using StringBuilder::insert can we just append `observation.hasComment() ? observation.getComment() : StringUtils.EMPTY` as the very last step of this method?

Have a look at what insert does. Every time we call it the entire string is shifted (copied) to the right by N places. This will not be very performant.

service/src/test/resources/ehr/mapper/observation/expected-output-observation-statement-25.xml
z-uddin: For Observation.component we handle the value[x] field as same for Observation (uncategorised).
And for this translation, have translated to valueCodeableConcept to Code Value string.
The double space is due to how PertinentInformationObservationMapper.mapObservationValueToPertinentInformation builds strings - using spaces. 
I did not write this method, so think the easiest solution is to call trim() after to avoid the double space if you agree?

service/src/test/resources/ehr/mapper/observation/expected-output-observation-statement-22.xml
z-uddin: Believe this being caused by the call to PertinentInformationObservationMapper.mapObservationValueToPertinentInformation, see comment above

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/PertinentInformationObservationValueMapper.java
```diff
@@ -200,4 +215,137 @@ private static boolean isDenominatorPresent(Ratio ratio) {
...
+                var lowValue = referenceRange.getLow().getValue().toString();
+                var lowUnit = referenceRange.getLow().getUnit();
+                var highValue = referenceRange.getHigh().getValue().toString();
+                var highUnit = referenceRange.getHigh().getUnit();
+
+                return String.format(COMPONENT_REFERENCE_RANGE_TEMPLATE, lowValue, lowUnit, highValue, highUnit);
+            }
+        }
+        return StringUtils.EMPTY;
+    }
```
andrewwa-kainos, line 307: Can't reference ranges have just a minimum, or just a maximum?

z-uddin: It's not explicitly stated in the mapping spreadsheet, will confirm with @dmcavenue1

dmcavenue1: Mapping sheet is slightly abridged for Observation.component because the attributes that can appear in component are the same types that appear elsewhere in Observation - an instance of Observation.component.referenenceRange is identical in structure to Observation.referenceRange and can have both low and high attributes.

z-uddin: Thanks for confirming. 
Currently when translating Observation.referenceRange we check that there is low value with unit and high value with unit for it to be translated. Thus believe component.referenceRange performs these same checks here.

andrewwa-kainos: In which case we need a decision on how to output it when only one is given. Currently:
> "Range: `minVal` `minUnit` `maxVal` `maxUnit`"

e.g.
> "Range: 0.1 mg/L 0.6 mg/L"

How about "Range: min `minVal` `minUnit` max `maxVal` `maxUnit`", where each part can be omitted?
both:
> "Range: min 0.1 mg/L max 0.6 mg/L"

only min:
> "Range: min 0.1 mg/L"

only max:
> "Range: max 0.6 mg/L"

z-uddin: Implemented suggested change. 
Reflected in both Observation.component.referenceRange and Observation(uncategorised).value when value is a valueRange - to match mapping spreadsheet.

service/src/test/resources/ehr/mapper/observation/example-observation-resource-26.json
```diff
@@ -52,6 +52,28 @@
...
+        },
+        {
+            "referenceRange": [
+                {
+                    "text": "High only",
+                    "low": {
+                        "value": 0.6,
+                        "unit": "mg/L"
+                    }
+                }
```
andrewwa-kainos, line 75: If these ranges are distinct from the other one above, might be nice to use different values so it's clear they're about different things.

Now we get
> [... Range: Low 0.1 mg/L High 0.6 mg/L] [Range: Low 0.1 mg/L] [Range: Low 0.6 mg/L]

which looks like repetition. If they're different, let's have the test data different. If they're the same, is the repetition appropriate?

z-uddin: Yep these ranges are distinct from each other. Will update the values to to make the data more distinct.

andrewwa-kainos (approved):
LGTM

mattd-kainos (approved):

PR #182, opened 2021-04-27
NIAD-1268: Update opentest variables

Broom94:


docker/vars.opentest.sh
```diff
@@ -7,11 +7,6 @@ export GP2GP_MONGO_DATABASE_NAME="gp2gp"
 export GP2GP_MHS_OUTBOUND_URL="http://mock-mhs-adaptor:8081/mock-mhs-endpoint"
 export GP2GP_GPC_GET_URL="http://gpcc:8090/B82617/STU3/1/gpconnect"
 
-export GP2GP_SPINE_CLIENT_CERT=""
```
n-r-b, line 10: If we're renaming these, should update the `env.opentest.example.yml` too with these changes?

mattd-kainos (approved):

n-r-b (approved):

PR #192, opened 2021-04-29
NIAD-1363: Fix order of Immunization ObservationStatement elements

z-uddin:
* Re-ordered Immunization ObservationStatement element order to follow sequence outlined on ticket.
* Updated test files accordingly.
* Renamed template file to match filenames of other templates.


service/src/main/resources/templates/ehr_immunization_observation_statement_template.mustache
```diff
@@ -14,9 +15,8 @@
             </pertinentAnnotation>
         </pertinentInformation>
         {{/pertinentInformation}}
-        {{{code}}}
         {{#participant}}
         {{{.}}}
         {{/participant}}
     </ObservationStatement>
-</component>
+</component>
```
andrewwa-kainos, line 22: It's often considered good practice to include a final new line. That's partly why github gives the little warning when it's absent. Most (possibly all?) of the reasons why it's good practice don't apply to us though. One effect we *can* see is the `</component><component>` on one line in the output, where we might prefer it to be formatted as it was before.

n-r-b (approved):

service/src/main/resources/templates/ehr_immunization_observation_statement_template.mustache
z-uddin: Added back final new line and updated tests 
Thanks for the suggestion 👍

Broom94 (approved):

andrewwa-kainos (approved):

PR #197, opened 2021-04-29
NIAD-1424 Remove assumption and add Mock MHS debug log messages

mattd-kainos:


mock-mhs-adaptor/src/main/java/uk/nhs/adaptors/mockmhsservice/service/MockMhsService.java
```diff
@@ -84,6 +84,7 @@
     }
 
     private void verifyOutboundMessagePayload(String requestBody) throws JsonProcessingException, MockMHSException {
+        LOGGER.debug("Received outbound MHS request payload:\n{}", requestBody);
```
andrewwa-kainos, line 87: One thing we could point out here is that there's some assumption about the platform, since you're using `\n`. I think we're happy with that, but if not:
```suggestion
        LOGGER.debug("Received outbound MHS request payload:{}{}", System.lineSeparator(), requestBody);
```

mock-mhs-adaptor/src/main/resources/application.yml
```diff
@@ -1,6 +1,13 @@
 server:
   port: ${MOCK_MHS_SERVER_PORT:8081}
 
+logging:
+  level:
+    uk:
+      nhs:
+        adaptors:
+          mockmhsservice: DEBUG
```
matthewdoreilly, line 9: Minor,  but I think you can just do

```
logging:
  level:
    root: DEBUG
```

ElliotB-96 (approved):

mock-mhs-adaptor/src/main/resources/application.yml
mattd-kainos: 👍  took the second option, we're not interested in debug for third party stuff

matthewdoreilly (approved):

Broom94 (approved):

andrewwa-kainos (approved):

PR #199, opened 2021-04-30
NIAD-1380 Encode the referenceRange text for XML

andrewwa-kainos:
Without encoding, range text such as "< 10" would be copied literally into the output XML,
making it badly-formed. Instead, this example should be "\&lt; 10"

I saw two immediate options:
1. To use apache commons escaping utils
2. To convert the string format constants into mustache templates, where escaping is given
 out of the box.

I opted for the latter for consistency with the rest of the codebase.

This included introducing a new way to create "ad hoc" Mustache objects that can be used
just like the other templates we're creating from files.

So the precedent we're setting here is to use the "ad hoc" approach when there's only one
variable and the whole template is small, otherwise make a file.


ElliotB-96 (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/StructuredObservationValueMapper.java
```diff
@@ -12,30 +13,24 @@
...
-        + "<originalText>%s</originalText>"
-        + "</interpretationCode>";
+    private static final Map<Class<? extends IBaseElement>, Function<IBaseElement, String>> VALUE_MAPPING_FUNCTIONS = Map.of(
+        Quantity.class, value -> ObservationValueQuantityMapper.processQuantity((Quantity) value),
+        StringType.class, value -> processStringType((StringType) value)
+    );
+    private static final Mustache STRING_VALUE_TEMPLATE = TemplateUtils.compileTemplate("<value xsi:type=\"ST\">{{value}}</value>");
+    private static final Mustache REF_RANGE_TEMPLATE = TemplateUtils.loadTemplate("ehr_reference_range_template.mustache");
+
+    private static final Mustache LOW_RANGE_TEMPLATE = TemplateUtils.compileTemplate("<low value=\"{{{low}}}\"/>");
```
n-r-b, line 31: Just a thought, we could just use a `.` here, then we don't have to create a Map to populate the value
```suggestion
    private static final Mustache LOW_RANGE_TEMPLATE = TemplateUtils.compileTemplate("<low value=\"{{{.}}}\"/>");
```

n-r-b (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/StructuredObservationValueMapper.java
andrewwa-kainos: Nice!

n-r-b (approved):

z-uddin (approved):

PR #202, opened 2021-05-04
NIAD-1365: Element order requeststatement

Broom94:


service/src/main/resources/templates/ehr_request_statement_template.mustache
```diff
@@ -1,26 +1,26 @@
...
         <responsibleParty typeCode="RESP">
             <agentRef classCode="AGNT">
                 <id root="{{{.}}}"/>
             </agentRef>
         </responsibleParty>
         {{/responsibleParty}}
+        {{#participant}}
+        {{{.}}}
+        {{/participant}}
+        {{#defaultReasonCode}}
```
z-uddin, line 22: Believe defaultReasonCode is used when code is not populated.
Thus think this should be moved to be above text element tag.

z-uddin (approved):

n-r-b (dismissed):

z-uddin (dismissed):

ElliotB-96 (approved):

z-uddin (approved):

ElliotB-96 (approved):

PR #218, opened 2021-05-07
NIAD-1426: Add agents to mapped procedure request

andrewwa-kainos:
PlanStatement/Participant maps to both author/agentRef and Participant2/@typeCode='PPRF'.
author/time holds the same value as effectiveTime


ElliotB-96 (approved):
Looks good.

msolecka (approved):

PR #229, opened 2021-05-18
NIAD-910: Fix issue where some derived observations were not being mapped properly

matthewdoreilly:
## Description

* Fix issue where some derived observations were not being mapped properly to the HL7/XML.
Correct logic is for each observation associated with the specimen, we retrieve the `observation.related` `has-member`  observations.
Previously, we were looking at the list of observation associated with the diagnostic report, and retrieving the `observation.related` `derived-from` observations, which wasn't correct.
* If an observation has other related observations (i.e., observations derive from it) then that observation's CompoundStatement's `classCode` should be `BATTERY`. Else, it should be `CLUSTER`

## Jira ticket
https://gpitbjss.atlassian.net/browse/NIAD-910


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -111,8 +114,16 @@ private String prepareCodeElement(Observation observation) {
...
+        if (observation.hasRelated()) {
+            return ClassCode.BATTERY;
+        }
+
+        return ClassCode.CLUSTER;
+    }
+
     private String prepareObservationStatement(IdMapper idMapper, Observation observation) {
-        if (observation.hasCode()) {
+        if (!CodeableConceptMappingUtils.hasCode(observation.getCode(), List.of(NARRATIVE_STATEMENT_CODE))) {
```
andrewwa-kainos, line 126: This looks like `CodeableConceptMappingUtils#hasCode` could use an overload accepting varargs. What do you think?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -264,6 +252,55 @@ private String prepareNarrativeStatement(Observation observation, String comment
...
     }
 
+    private String prepareStatementsForDerivedObservations(IdMapper idMapper, List<Observation> derivedObservations) {
+        StringBuilder derivedObservationsBlock = new StringBuilder();
+
+        derivedObservations.forEach(derivedObservation -> {
+            var observationStatement = prepareObservationStatement(idMapper, derivedObservation);
+            var narrativeStatement = prepareNarrativeStatement(idMapper, derivedObservation);
+
+            if (!observationStatement.isEmpty() && !narrativeStatement.isEmpty()) {
```
andrewwa-kainos, line 262: Should we also check for blankness?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -264,6 +252,55 @@ private String prepareNarrativeStatement(Observation observation, String comment
...
+            } else {
+                derivedObservationsBlock.append(observationStatement).append(narrativeStatement);
+            }
+        });
+
+        return derivedObservationsBlock.toString();
+    }
+
+    private CommentType prepareCommentType(Observation observation) {
+        if (observation.getComment().equals("EMPTY REPORT")) {
```
andrewwa-kainos, line 293: Is a comment guaranteed to be present or shall we yoda it?
```suggestion
        if ("EMPTY REPORT".equals(observation.getComment())) {
```

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapperTest.java
```diff
@@ -128,32 +120,31 @@ public void When_MappingObservationJson_Expect_CompoundStatementXmlOutput(String
...
-        );
-
-        String compoundStatementXml = observationMapper.mapObservationToCompoundStatement(
-            observationAssociatedWithSpecimen,
-            Collections.emptyList()
-        );
-
-        assertThat(compoundStatementXml).isEqualTo(expectedXmlOutput);
-    }
+    // TODO: Fix as part of NIAD-1469
```
andrewwa-kainos, line 129: 🚨

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -66,43 +69,43 @@
     private final ParticipantMapper participantMapper;
     private final RandomIdGeneratorService randomIdGeneratorService;
 
-    public String mapObservationToCompoundStatement(Observation observationAssociatedWithSpecimen, List<Observation> observations) {
+    public String mapObservationToCompoundStatement(Observation observationAssociatedWithSpecimen) {
```
andrewwa-kainos, line 72: With the amount of parameters being passed around for this now I'd be thinking about using an inner class to hold state. What do you think?

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapperTest.java
matthewdoreilly: This wouldn't be forgotten about (as https://gpitbjss.atlassian.net/browse/NIAD-1469 is a sub-task of NIAD-910).
So I thought fixing this test would make more sense for that task, but realised it's easy enough to fix this test as part of this branch.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -282,13 +253,39 @@ private String prepareStatementsForDerivedObservations(IdMapper idMapper, List<O
...
+                derivedObservationsBlock.append(observationStatement).append(narrativeStatements);
             }
         });
 
         return derivedObservationsBlock.toString();
     }
 
+    private boolean observationDoesNotHaveCommentNoteCode(Observation observation) {
+        return observation.hasCode() && !CodeableConceptMappingUtils.hasCode(observation.getCode(), List.of(COMMENT_NOTE_CODE));
+    }
```
andrewwa-kainos, line 265: Wouldn't `!observation.hasCode()` also satisfy `observationDoesNotHaveCommentNoteCode`?
```suggestion
    private boolean observationDoesNotHaveCommentNoteCode(Observation observation) {
        return !observation.hasCode() || !CodeableConceptMappingUtils.hasCode(observation.getCode(), List.of(COMMENT_NOTE_CODE));
    }
```

Or are we looking for something like `observationHasNonCommentNoteCode`?
```suggestion
    private boolean observationHasNonCommentNoteCode(Observation observation) {
        return observation.hasCode() && !CodeableConceptMappingUtils.hasCode(observation.getCode(), List.of(COMMENT_NOTE_CODE));
    }
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -282,13 +253,39 @@ private String prepareStatementsForDerivedObservations(IdMapper idMapper, List<O
...
+                    .filter(this::observationDoesNotHaveCommentNoteCode)
+                    .findFirst();
+
+            if (derivedObservationWithoutCommentNoteCode.isPresent()) {
+                return ClassCode.BATTERY;
+            }
+        }
+
+        return ClassCode.CLUSTER;
+    }
```
andrewwa-kainos, line 287: You can still call `.stream()` on an empty collection; `.findFirst()` will necessarily return an empty optional.
```suggestion
    private ClassCode prepareClassCode(List<Observation> derivedObservations) {
        return derivedObservations.stream()
            .filter(this::observationDoesNotHaveCommentNoteCode)
            .map($ -> ClassCode.BATTERY) // if present map to BATTERY else still empty stream
            .findFirst() // convert to optional
            .orElse(ClassCode.CLUSTER); // if empty use CLUSTER instead
    }
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
matthewdoreilly: It should only ever be empty. But wouldn't hurt checking for blankness since it shouldn't be outputted then obviously

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
matthewdoreilly: Good point, but we do a `observation.hasComment()` check beforehand so this is safe

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
matthewdoreilly: A number of places would need updated to make this change. Also since it'll then be treated as an array, we'd need to update the implementation of `CodeableConceptMappingUtils.hasCode` to check where the String equals "any String in array". I don't think this is worth the effort as part of this ticket

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
matthewdoreilly: I like `observationHasNonCommentNoteCode`, I'll change it to that

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
andrewwa-kainos: It adds a layer of abstraction (which may or may not be a good thing), and makes the processes within that layer a lot simpler - syntactically for sure, as the method signatures become cleaner, and imo in terms of readability too. You end up with a thing whose responsibility, and entire lifespan, is just processing a single mapping. The cost is in understanding that that is what is happening which could be confusing for readers who aren't familiar with the strategy, and... some extra indentation?

Both will work so I'm certainly not going to insist, but there's my suggestion! 😀

andrewwa-kainos (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ClassCode.java
```diff
@@ -0,0 +1,13 @@
+package uk.nhs.adaptors.gp2gp.ehr.mapper;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+public enum ClassCode {
```
dawidskainos, line 8: ClassCode is a rather generic name. Does it refer to CompundStatement only or is it common in all MIM elements?

matthewdoreilly: Related to `CompoundStatement` mainly, I can update that to make it clearer

dawidskainos (approved):

andrewwa-kainos (approved):

PR #232, opened 2021-05-19
Niad 1467 specimen narrative statement

dawidskainos:


andrewwa-kainos (approved):

service/src/main/resources/templates/specimen_compound_statement_template.mustache
```diff
@@ -23,14 +23,16 @@
                 </specimenSpecimenMaterial>
             </specimenRole>
         </specimen>
-        {{#pertinentInformation}}
-        <pertinentInformation typeCode="PERT">
-            <sequenceNumber value="+1" />
-            <pertinentAnnotation classCode="OBS" moodCode="EVN">
+        {{#narrativeStatement}}
```
matthewdoreilly, line 26: You could reuse the `narrative_statement_template.mustache` template for this, https://github.com/nhsconnect/integration-adaptor-gp2gp/blob/main/service/src/main/resources/templates/narrative_statement_template.mustache.

Then the mustache template would only need
```
{{#narrativeStatement}}
  {{{.}}}
{{/narrativeStatement}}
```

And then in `SpecimenCompoundStatementTemplateParameters`, you wouldn't need `private String narrativeStatementId;`

matthewdoreilly (approved):

PR #233, opened 2021-05-20
NIAD-1340: Agent directory based on actual agent references

msolecka:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectory.java
```diff
@@ -0,0 +1,106 @@
...
+        if (reference.hasReferenceElement()) {
+            String resourceType = reference.getReferenceElement().getResourceType();
+            if (ResourceType.Practitioner.name().equals(resourceType)) {
+                return putIfAbsent(buildAgentKeyForPractitioner(reference));
+            } else if (ResourceType.Organization.name().equals(resourceType)) {
+                return putIfAbsent(AgentKey.builder().organizationReference(reference.getReference()).build());
+            }
+        }
+
+        LOGGER.error("Invalid agent reference to not supported resource: " + reference.getReference());
```
mattd-kainos, line 42: Don't need to log and throw, can just throw with an appropriate message and let a downstream error handler log it out.

mattd-kainos:
There are multiple TODO comments in the code tagged NIAD-1340 that need to be resolved

mattd-kainos:
I added the following to IdMapper to see if any code was still trying to use IdMapper to agent-related references.

```
private static final Set<String> NOT_ALLOWED = Set.of(
        ResourceType.Organization.name(), 
        ResourceType.Practitioner.name(),
        ResourceType.PractitionerRole.name());
...
public String getOrNew(Reference reference, Boolean isResourceMapped) {
        if (NOT_ALLOWED.contains(reference.getReferenceElement().getResourceType())) {
            throw new RuntimeException("Not allowed to use agent-related resource with IdMapper");
        }
...
```

There are a few failing tests with this check. Most of them seem to be stale test setup, but at least one is in implementation: BloodPressureMapper

How do we feel about adding this check to IdMapper so we can fail-fast on any invalid usage?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectoryMapper.java
```diff
@@ -31,153 +26,57 @@
...
-        return observations.stream()
-            .map(observation -> buildAgentPerson(observation, bundle, mappedOrganizationsAndPractitioner))
-            .filter(StringUtils::isNotEmpty)
-            .collect(Collectors.toList());
-    }
 
-    private List<String> prepareReferralRequestAgentsList(Bundle bundle, Set<String> mappedOrganizationsAndPractitioner) {
-        List<ReferralRequest> referralRequests = AgentDirectoryExtractor.extractReferralRequestsWithRequester(bundle);
+        builder.agentPersons(
+            agentDirectories.stream()
```
mattd-kainos, line 46: I don't understand this code. Could we make it a bit more declarative?
Why is agentDirectories being filtered? I expect this to be a set of every agent that goes into the directory. Doesn't adding a filter step mean we're going to be missing agents that were referenced by our previously output statements?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectory.java
```diff
@@ -0,0 +1,106 @@
...
+    public Set<Map.Entry<AgentKey, String>> getEntries() {
+        return agentDirectory.entrySet();
+    }
+
+    public String getAgentId(Reference reference) {
+        if (reference.hasReferenceElement()) {
+            String resourceType = reference.getReferenceElement().getResourceType();
+            if (ResourceType.Practitioner.name().equals(resourceType)) {
+                return putIfAbsent(buildAgentKeyForPractitioner(reference));
+            } else if (ResourceType.Organization.name().equals(resourceType)) {
```
mattd-kainos, line 37: Can an agent reference be to an organisation only? I though it must always be to practitioner. @dmcavenue1 ?

msolecka: I believe agentOrganization related to managingOgranization of Patient

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectoryMapper.java
msolecka: The ides behind this is:
- first agent on agent directory is always with agentOrganization from patient.managingOrganization ( it is described in mapping excel, https://kainossoftwareltd-my.sharepoint.com/:x:/g/personal/davidmcav_kainos_com/Ef-K0P2Do9tMu13fI1gLkDsBES6fhhj8NN3pYzG_F3iJ5w?e=iiTKx9 on tab agentDirectory, at the top, also in ac of NIAd-1024
- then all other references go, so if other resource is referencing that organization also ( for example, observation.performer is sometimes referencing only organization) and in that case I want to map organization only once, so if it was mapped for agentOrganization first with the transformed id from AgenDirectory class, then it should not be mapped to normal agent again. 

But it is only applied when there is a single Organization, for Practitioner + Organization it will create agent with agentPerson and representedOrganization still.

What do you think about it now?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectory.java
msolecka: In public demonstrator, our test patient has Observations and DiagnosticReport that has single Organization reference in performer. Is it wrong?

msolecka: @mattd-kainos, Checked it with @dmcavenue1 and Observation and DiagnosticReport can have single Organization but it will be a corner case. And for now the mapping should behave as bellow:

- Organization referenced by Patient.managingOrganization is mapped once, and at first step as agentOrganization in agentDirectory
```
  <part typeCode="PART">
    <Agent classCode="AGNT">
        <id root="5E496953-065B-41F2-9577-BE8F2FBD0757" />
        <agentOrganization determinerCode="INSTANCE" classCode="ORG">
                        <id extension="A82038" root="2.16.840.1.113883.2.1.4.3"/>
            <name>TEMPLE SOWERBY MEDICAL PRACTICE</name>
            <telecom value="01133800000" use="WP"/>
            <addr use="WP">
                <streetAddressLine>Fulford Grange</streetAddressLine>
                <streetAddressLine>Micklefield Lane</streetAddressLine>
                <streetAddressLine>Rawdon</streetAddressLine>
                <streetAddressLine>Rawdon</streetAddressLine>
                <streetAddressLine>Leeds</streetAddressLine>
                <postalCode>LS19 6BA</postalCode>
            </addr>
        </agentOrganization>
    </Agent>
</part>
```
- then if any Observation or DiagnosticReport has reference to Organization only and that Organization is the same as in Patient.managingOrganization it will reference the one mapped to agentOrganization
- if any Observation or DiagnosticReport has reference to Organization only and that Organization is not in Patient.managingOrganization the agent is created with representedOrganization

```
<part typeCode="PART">
    <Agent classCode="AGNT">
        <id root="6D340A1B-BC15-4D4E-93CF-BBCB5B74DF73"/>
        <representedOrganization classCode="ORG" determinerCode="INSTANCE">
            <id extension="A82038" root="2.16.840.1.113883.2.1.4.3"/>
            <name>TEMPLE SOWERBY MEDICAL PRACTICE</name>
            <telecom value="01133800000" use="WP"/>
            <addr use="WP">
                <streetAddressLine>Fulford Grange</streetAddressLine>
                <streetAddressLine>Micklefield Lane</streetAddressLine>
                <streetAddressLine>Rawdon</streetAddressLine>
                <streetAddressLine>Rawdon</streetAddressLine>
                <streetAddressLine>Leeds</streetAddressLine>
                <postalCode>LS19 6BA</postalCode>
            </addr>
        </representedOrganization>
    </Agent>
</part>

```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectory.java
```diff
@@ -0,0 +1,102 @@
...
+
+        throw new EhrMapperException("Not supported agent reference to practitioner: " + practitionerReference.getReference()
+            + " and organization: " + organizationReference.getReference());
+    }
+
+    private AgentKey buildAgentKeyForPractitioner(Reference practitionerReference) {
+        return inputBundle.getEntry().stream()
+            .map(Bundle.BundleEntryComponent::getResource)
+            .filter(resource -> ResourceType.PractitionerRole.equals(resource.getResourceType()))
+            .map(resource -> (PractitionerRole) resource)
```
ElliotB-96, line 61: Optional thing but you could use PractitionerRole.class::cast

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectoryMapper.java
mattd-kainos: Yes that makes sense now, and I see there is a test that describes this behaviour also When_MappingAgentDirectoryWithPatientManagingOrganizationInAgentKeys_Expect_AgentPersonNotDuplicated. I should have looked a bit deeper. :thumbsup:

mattd-kainos:
I have verified that all UAT agentRef/id refer to actual Agent/id: https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/247

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectoryMapper.java
```diff
@@ -31,153 +26,57 @@
...
-            .map(Optional::get)
-            .map(Organization.class::cast)
-            .map(organizationToAgentMapper::mapOrganizationToAgent)
+        Optional<Organization> patientManagingOrganization = extractPatientManagingOrganization(bundle, nhsNumber);
+
+        patientManagingOrganization
+            .map(organization -> OrganizationToAgentMapper.mapOrganizationToAgent(organization, getAgentId(organization)))
+            .map(builder::patientManagingOrganization)
             .orElseThrow(() -> new EhrMapperException("The ASR bundle does not contain a Patient resource with the correct identifier "
                 + "and a managingOrganization"));
```
andrewwa-kainos, line 43: I found this a bit tricky to follow, I think because some of the ops work but aren't really thematic.

If I understand it right, the exception should be thrown if `patientManagingOrganization` comes back empty, rather than from the interim ops, so my first suggestion is to put the `.orElseThrow()` there. Then `patientManagingOrganization` is just an `Organization` rather than an `Optional` of one.

Then the builder can be invoked in a chain for the two properties, `.patientManagingOrganization()` and `.agentPersons()`.

```suggestion
        Organization patientManagingOrganization = extractPatientManagingOrganization(bundle, nhsNumber)
            .orElseThrow(() -> new EhrMapperException("The ASR bundle does not contain a Patient resource with the correct identifier "
                + "and a managingOrganization"));

        String agentId = getAgentId(patientManagingOrganization);
        builder.patientManagingOrganization(mapOrganizationToAgent(patientManagingOrganization, agentId))
            .agentPersons(agentDirectories.stream()
                .filter(agentDirectory -> !isPatientManagingOrganization(patientManagingOrganization, agentDirectory))
                .map(this::mapAgentDirectory)
                .collect(Collectors.toList())
            );
```

What do you think?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentPersonMapper.java
```diff
@@ -17,40 +17,58 @@
...
         var builder = PractitionerAgentPersonMapperParameters.builder()
-            .practitionerId(messageContext.getIdMapper().getOrNew(ResourceType.Practitioner, practitioner.getIdElement()));
+            .agentId(agentDirectoryId);
 
-        buildPractitionerPrefix(practitioner).ifPresent(builder::practitionerPrefix);
-        var practitionerGiven = buildPractitionerGivenName(practitioner);
-        var practitionerFamily = buildPractitionerFamilyName(practitioner);
+        Optional<Practitioner> practitioner = messageContext.getInputBundleHolder()
+            .getResource(new IdType(agentKey.getPractitionerReference()))
+            .map(resource -> (Practitioner) resource);
```
andrewwa-kainos, line 33: I find the cast method clearer, personally

```suggestion
            .map(Practitioner.class::cast);
```

What do you think?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentPersonMapper.java
```diff
@@ -17,40 +17,58 @@
...
+            }
         }
 
-        practitionerRole.flatMap(this::buildPractitionerRole).ifPresent(builder::practitionerRole);
-        organization.map(organizationToAgentMapper::mapOrganizationToAgentInner).ifPresent(builder::organization);
+        buildPractitionerRole(agentKey).ifPresent(builder::practitionerRole);
+
+        messageContext.getInputBundleHolder()
+            .getResource(new IdType(agentKey.getOrganizationReference()))
+            .map(resource -> (Organization) resource)
```
andrewwa-kainos, line 54: Again if you agree

```suggestion
            .map(Organization.class::cast)
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentPersonMapper.java
```diff
@@ -17,40 +17,58 @@
...
+            .getPractitionerRoleFor(
+                agentKey.getPractitionerReference(),
+                agentKey.getOrganizationReference()
+            );
+
+        if (practitionerRole.isPresent() && hasPractitionerRoleDisplay(practitionerRole.get())) {
+            return Optional.of(practitionerRole.get().getCodeFirstRep().getCodingFirstRep().getDisplay());
         }
+
         return Optional.empty();
```
andrewwa-kainos, line 72: This is equivalent to the following 
```suggestion
        return practitionerRole
            .filter(AgentPersonMapper::hasPractitionerRoleDisplay)
            .map(PractitionerRole::getCodeFirstRep)
            .map(CodeableConcept::getCodingFirstRep)
            .map(Coding::getDisplay);
```

If you think it's more readable

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapper.java
```diff
@@ -64,7 +64,7 @@ public String mapEncounterToEhrComposition(Encounter encounter) {
             .originalText(buildOriginalText(encounter));
 
         final String recReference = findParticipantWithCoding(encounter, ParticipantCoding.RECORDER)
-            .map(idMapper::get)
+            .map(messageContext.getAgentDirectory()::getAgentId)
```
andrewwa-kainos, line 67: Could consider extracting `messageContext.getAgentDirectory()` to a variable, since it's called twice in this method.

It can be important to note that if the thing on the left of the `::` in a method reference needs evaluating, it is evaluated regardless of whether or not the lambda is actually invoked. In this case that means that it really is a refactoring and we're not adding any calls that wouldn't be made anyway.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -31,6 +36,10 @@ public String getOrNew(Reference reference) {
     }
 
     public String getOrNew(Reference reference, Boolean isResourceMapped) {
+        if (NOT_ALLOWED.contains(reference.getReferenceElement().getResourceType())) {
+            throw new RuntimeException("Not allowed to use agent-related resource with IdMapper");
```
andrewwa-kainos, line 40: Is there a more specific exception we could throw here?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -52,33 +61,14 @@ public boolean hasIdBeenMapped(Reference reference) {
...
-            return entry.getValue().getId();
-        }
-
-        LOGGER.warn("Unable to find a replacement resource for {}", referenceValue);
-        return null;
-    }
-
-    public String get(ResourceType resourceType, IdType id) throws EhrMapperException {
-        return get(buildReference(resourceType, id));
+        throw new EhrMapperException("Resource referenced was not mapped" + referenceValue);
```
andrewwa-kainos, line 71: May need a space
```suggestion
        throw new EhrMapperException("Resource referenced was not mapped " + referenceValue);
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -75,23 +76,46 @@ public String mapObservationToObservationStatement(Observation observation, bool
...
+            participantReference = messageContext.getAgentDirectory().getAgentRef(practitionerReference.get(), organizationReference.get());
+        } else if (practitionerReference.isPresent()) {
+            participantReference = messageContext.getAgentDirectory().getAgentId(practitionerReference.get());
+        } else if (organizationReference.isPresent()) {
+            participantReference = messageContext.getAgentDirectory().getAgentId(organizationReference.get());
+        } else {
+            throw new EhrMapperException(
+                "Invalid performer reference in observation resource with id: " + observation.getIdElement().getIdPart()
+            );
+        }
```
andrewwa-kainos, line 107: This might be a little easier to follow
```suggestion
        String participantReference;
        if (practitionerReference.isPresent() && organizationReference.isPresent()) {
            participantReference = messageContext.getAgentDirectory().getAgentRef(practitionerReference.get(), organizationReference.get());
        } else {
            var ref = practitionerReference.orElseGet(() -> organizationReference
                .orElseThrow(() -> new EhrMapperException("Invalid performer reference in observation resource with id: "
                    + observation.getIdElement().getIdPart())));
            participantReference = messageContext.getAgentDirectory().getAgentId(ref);
        }
```
Mostly I like it more because it doesn't use `Optional::get`, which always makes me nervous. What do you think?

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectoryMapperTest.java
```diff
@@ -1,82 +1,150 @@
...
-        var expectedOutput = ResourceTestFileUtils.getFileContent(DIRECTORY_NO_ROLES);
+    public void When_MappingAgentDirectoryWithoutPatientManagingOrganizationReference_Expect_Exception() throws IOException {
+        var jsonInput = ResourceTestFileUtils.getFileContent(INPUT_AGENT_DIRECTORY_WITHOUT_MANAGING_ORGANIZATION_REFERENCE);
+        Bundle bundle = fhirParseService.parseResource(jsonInput, Bundle.class);
+        initializeMessageContextWithAgentKeys(bundle);
+
+        assertThrows(
+            EhrMapperException.class,
+            () -> agentDirectoryMapper.mapEHRFolderToAgentDirectory(bundle, NHS_NUMBER)
+        );
```
andrewwa-kainos, line 92: If we're using assertj elsewhere can we stick to assertj here too?

```suggestion
        assertThatThrownBy(() -> agentDirectoryMapper.mapEHRFolderToAgentDirectory(bundle, NHS_NUMBER))
            .isExactlyInstanceOf(EhrMapperException.class);
            // could also include .hasMessage("whatever we expect here")
```

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
```diff
@@ -202,4 +205,9 @@ private Encounter extractEncounter(Bundle bundle) {
             Arguments.of(INPUT_BUNDLE_WITH_LIST_NOT_IN_TOPIC_OR_CATEGORY)
         );
     }
+
+    private void setUpMock() {
```
andrewwa-kainos, line 209: You could consider using a `@NestedClass` to group tests by mocking requirements rather than repeating this method invocation. Is there a common category for all these tests that need `setUpMock()`?

andrewwa-kainos (changes requested):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectory.java
```diff
@@ -0,0 +1,102 @@
...
+            .map(PractitionerRole.class::cast)
+            .filter(resource -> resource.getPractitioner().hasReference())
+            .filter(resource -> StringUtils.equals(resource.getPractitioner().getReference(), practitionerReference.getReference()))
+            .filter(resource -> resource.getOrganization().hasReference())
+            .map(resource -> AgentKey.builder()
+                .practitionerReference(resource.getPractitioner().getReference())
+                .organizationReference(resource.getOrganization().getReference())
+                .build())
+            .findFirst()
+            .orElseGet(() -> AgentKey.builder().practitionerReference(practitionerReference.getReference()).build());
```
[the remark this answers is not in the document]
msolecka: I think we have, it is private method used by getAgentId method only, and there is a check if hasReferenceElement, so fo sure it's not empty at this point.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
msolecka: the name for the method is wrong, I think it should be rather initializeMessageContext because it requires message context to be set up with different input bundle for each test. I will change it a little, I am not sure how to use nested class in this case. Let me know what you think or how you would like it to look.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
msolecka: I think it's better thanks

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AgentDirectoryMapper.java
msolecka: you're right, I changes it a little different, please take a look.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
andrewwa-kainos: With your latest changes, a `@NestedClass` would no longer work. 👍

andrewwa-kainos (approved):

Broom94 (approved):

mattd-kainos (approved):

PR #234, opened 2021-05-20
NIAD-1378 Ignore empty Encounter elements when mapping EHR extract

lgonczar:
... so that they don't end up in the output as ehrComposition elements without any components.

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
@@ -1,14 +1,17 @@
 package uk.nhs.adaptors.gp2gp.ehr.mapper;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import com.github.mustachejava.Mustache;
 
 import lombok.RequiredArgsConstructor;
 
+import lombok.extern.slf4j.Slf4j;
 import org.hl7.fhir.dstu3.model.Bundle;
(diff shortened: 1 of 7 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EhrExtractMapper.java
```diff
@@ -23,8 +26,10 @@
 
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 @Component
+@Slf4j
 public class EhrExtractMapper {
     private static final Mustache EHR_EXTRACT_TEMPLATE = TemplateUtils.loadTemplate("ehr_extract_template.mustache");
+    private static final String CONSULTATION_LIST_CODE = "325851000000107";
```
lgonczar, line 32: In a future PR I'm going to extracted codes like this out of individual classes into a common place.

service/src/test/resources/ehr/request/fhir/output/expected-xml-with-effective-time.xml
```diff
@@ -27,7 +27,7 @@
             <id root="test-id-2" />
             <statusCode code="COMPLETE" />
             <effectiveTime>
-                <low value="20100113"/><high value="20130713162000"/>
+                <low value="20100113"/>
```
lgonczar, line 30: High value disappeared because the corresponding ehrComposition doesn't exist in this output anymore.

dawidskainos (approved):

Broom94 (approved):

PR #239, opened 2021-05-24
NIAD-1470: If Diagnostic Report has no specimens, create a dummy/default specimen, and associate observations with that specimen.

matthewdoreilly:
Jira ticket: https://gpitbjss.atlassian.net/browse/NIAD-1470

* If Diagnostic Report has no specimens, create a dummy/default specimen, and then associate the observations with that specimen
(Previously we were creating the dummy/default specimen, but the observations were not being associated with it)
* Minor change to update `Specimen/Default-1` to `Specimen/DUMMY-SPECIMEN`, and `Observation/Default-1` to `Observation/DUMMY-OBSERVATION` (better readability)
* Remove `interpretationCode` from `CompoundStatement` level output (as it invalidates schema, and will be represented as a `NarrativeStatement`.
* `Participant` should be outputted to the bottom of HL7/XML.


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -294,7 +296,7 @@ private CompoundStatementClassCode prepareClassCode(List<Observation> derivedObs
         }
 
         private CommentType prepareCommentType(Observation observation) {
-            if (observation.getComment().equals("EMPTY REPORT")) {
+            if (observation.getId().equals(DUMMY_OBSERVATION_ID)) {
```
matthewdoreilly, line 299: It's safer to check the ID here (than check the comment which is user entered/free text)

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/SpecimenMapperTest.java
```diff
@@ -111,7 +111,7 @@ public void When_MappingDefaultSpecimenJson_Expect_DefaultSpecimenStatementXmlOu
             TEST_FILE_DIRECTORY + "expected_output_default_specimen.xml"
         );
 
-        Observation observation = new Observation().setSpecimen(new Reference().setReference("Specimen/Default-1"));
+        Observation observation = new Observation().setSpecimen(new Reference().setReference("Specimen/DUMMY-SPECIMEN"));
```
msolecka, line 114: I think, It could be empty observation here, without specimen reference, because in case the specimen is default, all observations not related to dummy specimen are used. So the reference should not be used in this case.

matthewdoreilly: This test is testing for a default specimen (i.e. no specimens) and a default observation (i.e. no observations).
I've added another test (`When_MappingDefaultSpecimenWithObservations_Expect_DefaultSpecimenAndObservationsXmlOutput`) for a default specimen (i.e. no specimens) but there is observations)

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/DiagnosticReportMapper.java
```diff
@@ -41,6 +41,9 @@
 @Slf4j
 public class DiagnosticReportMapper {
 
+    public static final String DUMMY_SPECIMEN_ID = "Specimen/DUMMY-SPECIMEN";
```
mattd-kainos, line 44: 1) This format "ResourceType/Id" is a reference not an ID
2) I'm pretty sure we can't use hardcoded references here. What happens when the patient record contains multiple diagnostic reports with no specimen?

matthewdoreilly: > This format "ResourceType/Id" is a reference not an ID

Updated now to just "DUMMY-SPECIMEN" and "DUMMY-OBSERVATION".

> What happens when the patient record contains multiple diagnostic reports with no specimen?

That's fine, this ID is only used in our code for the dummy/default Specimen generation (to keep the schema valid, and checks similar to this, https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/239/files#diff-60fce59b547f21c0c6a2664153b1a9f2d3e2e3432d3eb8131b10ae658ed43fe6R299).
In the HL7/XML output, the ID of a dummy/default Specimen won't be "DUMMY-SPECIMEN", it'll be a randomly generated UUID, like the example below:
```
<component typeCode="COMP" contextConductionInd="true">
    <CompoundStatement classCode="CLUSTER" moodCode="EVN">
        <id root="3316531F-5705-424C-9E1A-EE694FB411B4"/>
        <code code="123038009" codeSystem="2.16.840.1.113883.2.1.3.2.4.15" displayName="specimen (specimen)"/>
        <statusCode code="COMPLETE"/>
        <effectiveTime>
            <center nullFlavor="NI"/>
        </effectiveTime>
        <availabilityTime value="20120726134600"/>
        <specimen typeCode="SPC">
            <specimenRole classCode="SPEC">
                <id root="1141EEE6-EDE3-4358-8554-FFED0EE59284"/>
                <id root="2.16.840.1.113883.2.1.4.5.2" extension="DUMMY"/>
                <effectiveTime>
                    <center value="20120726134600"/>
                </effectiveTime>
                <specimenSpecimenMaterial determinerCode="INSTANCE" classCode="MAT">
                    <desc>UNKNOWN</desc>
                </specimenSpecimenMaterial>
            </specimenRole>
        </specimen>
```

So if we have multiple diagnostic reports with no specimen, they'll always have different UUIDs (no conflicts)

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -95,6 +94,8 @@ public String map() {
                 .collect(Collectors.toList());
 
             var classCode = prepareClassCode(derivedObservations);
+            var compoundStatementId = idMapper.getOrNew(ResourceType.Observation, observationAssociatedWithSpecimen.getIdElement());
```
msolecka, line 97: this line will generate same id for each default observation if only there will be more default observations within on bundle. That will happen because idMapper store map of ids and its mappings and all default observations has the same id so mapped id will also be the same. Because the mapping of the ids must be consistent. 

We could try to add something random to the default observation id, that it would generate new id each time.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/DiagnosticReportMapper.java
```diff
@@ -180,7 +182,7 @@ private String buildNarrativeStatementForDiagnosticReport(DiagnosticReport diagn
         var narrativeStatementTemplateParameters = NarrativeStatementTemplateParameters.builder()
             .narrativeStatementId(randomIdGeneratorService.createNewId())
             .commentType(commentType)
-            .issuedDate(DateFormatUtil.toHl7Format(diagnosticReport.getIssued().toInstant()))
+            .commentDate(DateFormatUtil.toHl7Format(diagnosticReport.getIssued().toInstant()))
```
mattd-kainos, line 185: We can't use .toInstant() from the HAPI FHIR library because of variable-precision dates/time e.g. if there is only a date we can't assume the time is midnight. There is another DateFormatUtil method that support the FHIR *Element class instead.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -225,7 +180,7 @@ private String mapObservationToNarrativeStatement(Observation observation, Strin
             var narrativeStatementTemplateParameters = NarrativeStatementTemplateParameters.builder()
                 .narrativeStatementId(randomIdGeneratorService.createNewId())
                 .commentType(commentType)
-                .issuedDate(DateFormatUtil.toHl7Format(observation.getIssued().toInstant()))
+                .commentDate(DateFormatUtil.toHl7Format(observation.getIssued().toInstant()))
```
mattd-kainos, line 183: as above about not using toInstant

msolecka (approved):

mattd-kainos (approved):

PR #241, opened 2021-05-24
NIAD-1449: Dynamic test fixture to transform arbitrary json ASR payload files

AlanSimpson1723:


README.md
```diff
@@ -285,6 +285,28 @@ We provide a mock MHS adaptor for local development and testing.
...
+## How to transform arbitrary json ASR payload files
+
+This is a document on 
+how to run tests on json files to convert them into the Xml Files for analysis
+
+####Step 1.
+
+Please locate the file Input folder \
+("src/test/resources/JsonToXml/Input/") \
+Place all Json files you want need to convert in here
```
msolecka, line 297:
```suggestion
Place all Json files to convert in here
```

README.md
```diff
@@ -285,6 +285,28 @@ We provide a mock MHS adaptor for local development and testing.
...
+####Step 2.
+
+Please locate the TransformJsonToXml.java class\
+("src/test/java/uk.nhs.adaptors.gp2gp/")\
+Run the TransformJsonToXml.java main method
+
+####Step 3.
+
+The Converted .Xml files will be located in
+("src/test/resources/JsonToXml/Output/")
```
msolecka, line 308: I think we have the convention to name folders/packages with all lowercase. It would be good to follow it here. https://gpitbjss.atlassian.net/wiki/spaces/NIA/pages/2108522539/Java+Coding+Standards

README.md
```diff
@@ -285,6 +285,28 @@ We provide a mock MHS adaptor for local development and testing.
 | MOCK_MHS_SERVER_PORT                 | 8081                      | The port on which the mock MHS Adapter will run.
 | MOCK_MHS_LOGGING_LEVEL               | INFO                      | Mock MHS logging level. One of: DEBUG, INFO, WARN, ERROR. The level DEBUG **MUST NOT** be used when handling live patient data.
 
+## How to transform arbitrary json ASR payload files
+
+This is a document on 
+how to run tests on json files to convert them into the Xml Files for analysis
```
bartek-sarul, line 291: This will be properly displayed but it's just a bit strange when looking at source code

README.md
```diff
@@ -285,6 +285,28 @@ We provide a mock MHS adaptor for local development and testing.
...
+
+Please locate the file Input folder \
+("src/test/resources/JsonToXml/Input/") \
+Place all Json files you want need to convert in here
+
+####Step 2.
+
+Please locate the TransformJsonToXml.java class\
+("src/test/java/uk.nhs.adaptors.gp2gp/")\
+Run the TransformJsonToXml.java main method
```
bartek-sarul, line 303: Add a bash file that runs this

service/src/test/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,209 @@
...
+        List<String> fileNames = new ArrayList<>();
+
+        for (File jsonFile : files) {
+            if (FilenameUtils.getExtension(jsonFile.getName()).equalsIgnoreCase("json")) {
+                try {
+                    String jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + jsonFile.getName());
+                    jsonStringInputs.add(jsonAsString);
+                    fileNames.add(jsonFile.getName());
+                } catch (Exception e) {
+                    e.printStackTrace();
```
bartek-sarul, line 85: If it breaks, it might be worth displaying on which file it failed

service/src/test/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+        }
+        return InputWrapper.builder()
+                .jsonFileInputs(jsonStringInputs)
```
bartek-sarul, line 90: Not our coding style. Remove indentation

service/src/test/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,209 @@
...
+        return nhsNumber;
+    }
+
+    private static String readJsonFileAsString(String file) throws Exception {
+        return new String(Files.readAllBytes(Paths.get(file)));
+    }
+
+    private static void writeToFile(String xml, String fileName) throws IOException {
+        String fileOutputName = FilenameUtils.removeExtension(fileName);
+        BufferedWriter writer = new BufferedWriter(new FileWriter(XML_OUTPUT_PATH + fileOutputName + ".xml"));
```
bartek-sarul, line 112: Use "try with resources"

bartek-sarul (changes requested):

service/src/test/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,209 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransformJsonToXml {
+
+    private static final String JSON_FILE_INPUT_PATH =
+            Paths.get("src", "test", "resources", "JsonToXml", "Input").toFile().getAbsoluteFile().getAbsolutePath() + "/";
+    private static final String XML_OUTPUT_PATH =
+            Paths.get("src", "test", "resources", "JsonToXml", "Output").toFile().getAbsoluteFile().getAbsolutePath() + "/";
```
mattd-kainos, line 58: The code is not loading these as classpath resources, and writing out files into src/test/resources can cause some IDE havok

Should we load/write these relative to the working directory e.g. `test_fixtures/input` and `test_fixtures/output` so they are not in our source tree?

We should also add these paths to `.gitignore` so the files are not accidentally committed.

service/src/test/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+        }
+        return InputWrapper.builder()
+                .jsonFileInputs(jsonStringInputs)
+                .jsonFileNames(fileNames)
+                .build();
+    }
+
+    private static String stringToJson(String json) throws JSONException {
```
msolecka, line 95: I am not sure we can assume patient resource or nhs identifier will be first in the list always. It would be good to parse the resource to fhir bundle and filter patient resource and valid identifier

something like that, with usage of FhirParseService we have already, you can search for it in other tests.
```
 Bundle bundle = new FhirParseService().parseResource(json, Bundle.class);
        return bundle.getEntry().stream()
            .map(Bundle.BundleEntryComponent::getResource)
            .filter(resource -> ResourceType.Patient.equals(resource.getResourceType()))
            .map(resource -> (Patient)resource)
            .map(resource -> getNhsNumberIdentifier(nhsnumberSystem, resource))
            .filter(Optional::isPresent)
            .map(identifier -> identifier.get().getValue())
            .findFirst()
            .orElseThrow(() -> new InvalidJsonException("No nhs number in Patient resource"));
```
The method could be called extractNhsNumber or getNhsNumber to reflect what it is doing.

service/src/test/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,209 @@
...
+                nonConsultationResourceMapper, agentDirectoryMapper, messageContext);
+
+
+        final EhrExtractTemplateParameters ehrExtractTemplateParameters =
+                ehrExtractMapper.mapBundleToEhrFhirExtractParams(getGpcStructuredTaskDefinition, bundle);
+
+        final String ehrExtractContent = ehrExtractMapper.mapEhrExtractToXml(ehrExtractTemplateParameters);
+
+        final String hl7TranslatedResponse = outputMessageWrapperMapper.map(getGpcStructuredTaskDefinition, ehrExtractContent);
+
```
msolecka, line 197: messageContext.resetMessageContext() should be executed here otherwise in case there will be more files to parse the ids and will shared between them and it will produce wrong output.

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+
+public class TransformJsonToXml {
+
+    private static final String JSON_FILE_INPUT_PATH =
+            Paths.get("src").toFile().getAbsoluteFile().getAbsolutePath() + "/../../transformJsonToXml/input/";
+    private static final String XML_OUTPUT_PATH =
+            Paths.get("src").toFile().getAbsoluteFile().getAbsolutePath() + "/../../transformJsonToXml/output/";
```
lgonczar, line 58: How about making the class accept these two paths as arguments, instead? Those arguments would be passed by the script/command that runs the app.

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+
+        for (File jsonFile : files) {
+            LOGGER.info("Parsing File: " + jsonFile.getName());
+            if (FilenameUtils.getExtension(jsonFile.getName()).equalsIgnoreCase("json")) {
+                try {
+                    String jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + jsonFile.getName());
+                    jsonStringInputs.add(jsonAsString);
+                    fileNames.add(jsonFile.getName());
+                } catch (Exception e) {
+                    e.printStackTrace();
```
lgonczar, line 89: Best to log an error using `LOGGER`, with a message that would provide some context and `e` as the second argument.

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+
+    public static void main(String[] args) throws Exception {
+
+        var inputWrapper = getFiles();
+
+        for (int i = 0; i < inputWrapper.getJsonFileInputs().size(); i++) {
+            var jsonString = inputWrapper.getJsonFileInputs().get(i);
+            String xmlResult = mapJsonToXml(jsonString);
+            var fileName = inputWrapper.getJsonFileNames().get(i);
+            writeToFile(xmlResult, fileName);
```
lgonczar, line 71: Maybe it would be worth replacing the extension to `xml` in the output file?

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+    private static String readJsonFileAsString(String file) throws Exception {
+        return new String(Files.readAllBytes(Paths.get(file)));
+    }
+
+    private static void writeToFile(String xml, String sourceFileName) {
+        String outputFileName = FilenameUtils.removeExtension(sourceFileName);
+        try {
+            BufferedWriter writer = new BufferedWriter(new FileWriter(XML_OUTPUT_PATH + outputFileName + ".xml"));
+            writer.write(xml);
+            writer.close();
```
mattd-kainos, line 124: Will this file be closed if an exception is thrown? Close always needs to be in a finally {} clause. You can use a try-with-resources to avoid needing to do it yourself
```suggestion
        try (BufferedWriter writer = new BufferedWriter(new FileWriter(XML_OUTPUT_PATH + outputFileName + ".xml"))) {
            writer.write(xml);
```

is the same as:

```
try {
            BufferedWriter writer = new BufferedWriter(new FileWriter(XML_OUTPUT_PATH + outputFileName + ".xml"));
            writer.write(xml);
            ...
} finally {
          writer.close();
}

File: service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java [hunk 15]
```diff
@@ -0,0 +1,225 @@
...
+            LOGGER.info("Contents of file: " + sourceFileName + ". Saved to: " + outputFileName + ".xml");
+        } catch (IOException e) {
+            LOGGER.info("Could not send Xml result to the file");
+        }
+    }
+
+    private static String mapJsonToXml(String jsonAsStringInput) {
```
mattd-kainos, line 131: This code is nearly the same as EhrExtractUATTest::setUp. Could they be refactored into a common class instead of duplicating them? I think the only differences would be a few dependencies:

| TransformJsonToXml | EhrExtractUATTest |
|----|----|
| RandomIdGeneratorService - random ids | RandomIdGeneratorService**Stub** |
| TimestampService - actual "now" timestamps | TimestampService - mocked constant timestamp |

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+    public static void main(String[] args) throws Exception {
+
+        var inputWrapper = getFiles();
+
+        for (int i = 0; i < inputWrapper.getJsonFileInputs().size(); i++) {
+            var jsonString = inputWrapper.getJsonFileInputs().get(i);
+            String xmlResult = mapJsonToXml(jsonString);
+            var fileName = inputWrapper.getJsonFileNames().get(i);
+            writeToFile(xmlResult, fileName);
+        }
```
andrewwa-kainos, line 72: I suggest extracting the lists so you don't have to get them new each time they're needed
```suggestion
        var jsonFileInputs = inputWrapper.getJsonFileInputs();
        var jsonFileNames = inputWrapper.getJsonFileNames();
        for (int i = 0; i < jsonFileInputs.size(); i++) {
            String jsonString = jsonFileInputs.get(i);
            String xmlResult = mapJsonToXml(jsonString);
            String fileName = jsonFileNames.get(i);
            writeToFile(xmlResult, fileName);
        }
```

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+                    String jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + inputJsonFile.getName());
+                    jsonStringInputs.add(jsonAsString);
+                    fileNames.add(inputJsonFile.getName());
+                } catch (Exception e) {
+                    LOGGER.info("Could Not Read Json Files.");
+                }
+            } else {
+                LOGGER.info("No .json Files Found");
+            }
+        }
```
andrewwa-kainos, line 94: This message is a little misleading - it will be printed every time a file is found that doesn't have a json extension. So if you had a directory of 2 .json files and 3 .xml files, it would print that not json files were found, 3 times!

I suggest detecting if the array is empty before looping. You might also consider that to be an exception scenario.

Then you can loop with more idiomatic Java 8+, especially if you add a lombok `@SneakyThrows` on `readJsonFileAsString()`

```suggestion
        if (files.length == 0) {
            throw new Exception("No json files found");
        }
        Arrays.stream(files)
            .peek(file -> LOGGER.info("Parsing file: {}", file.getName()))
            .filter(file -> FilenameUtils.getExtension(file.getName()).equalsIgnoreCase("json"))
            .forEach(file -> {
                String jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + inputJsonFile.getName());
                jsonStringInputs.add(jsonAsString);
                fileNames.add(inputJsonFile.getName());
            });
```

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    private static String extractNhsNumber(String json) {
+        var nhsnumberSystem = "https://fhir.nhs.uk/Id/nhs-number";
+        var bundle = FHIR_PARSE_SERVICE.parseResource(json, Bundle.class);
+        var nhsNumber = bundle.getEntry().stream()
+                .map(Bundle.BundleEntryComponent::getResource)
+                .filter(resource -> ResourceType.Patient.equals(resource.getResourceType()))
+                .map(resource -> (Patient) resource)
+                .map(resource -> (Identifier) getNhsNumberIdentifier(nhsnumberSystem, resource))
```
andrewwa-kainos, line 105:
```suggestion
                .map(Patient.class::cast)
                .map(resource -> getNhsNumberIdentifier(nhsnumberSystem, resource))
                .map(Identifier.class::cast)
```

If `getNhsNumberIdentifier()` always returns an `Identifier` we should cast it there and return it as the known type.

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+                .findFirst().get().getValue();
```
andrewwa-kainos, line 106: If the stream is empty then `findFirst()` will return an empty Optional and `.get()` will throw an exception. You might prefer to write your own exception so that it's understood that that is what you intended

```suggestion
                .findFirst()
                .elseThrow(() -> new SomeKindOfException("meaningful message"))
                .getValue();
```

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+                }
+            } else {
+                LOGGER.info("No .json Files Found");
+            }
+        }
+        return InputWrapper.builder().jsonFileInputs(jsonStringInputs).jsonFileNames(fileNames).build();
+    }
+
+    private static String extractNhsNumber(String json) {
+        var nhsnumberSystem = "https://fhir.nhs.uk/Id/nhs-number";
```
andrewwa-kainos, line 99: Could be nice to stick to camelCase
```suggestion
        var nhsNumberSystem = "https://fhir.nhs.uk/Id/nhs-number";
```

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+        var nhsNumber = bundle.getEntry().stream()
+                .map(Bundle.BundleEntryComponent::getResource)
+                .filter(resource -> ResourceType.Patient.equals(resource.getResourceType()))
+                .map(resource -> (Patient) resource)
+                .map(resource -> (Identifier) getNhsNumberIdentifier(nhsnumberSystem, resource))
+                .findFirst().get().getValue();
+        return nhsNumber;
+    }
+
+    private static Object getNhsNumberIdentifier(String nhsnumberSystem, Patient resource) {
```
andrewwa-kainos, line 110: And here, re: camelCase
```suggestion
    private static Object getNhsNumberIdentifier(String nhsNumberSystem, Patient resource) {
```

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+        return resource.getIdentifier()
+                .stream().filter(identifier -> identifier.getSystem().equals(nhsnumberSystem)).findFirst().get();
```
andrewwa-kainos, line 112: We should consider what happens if the identifier isn't found; i.e., `findFirst()` returns an empty Optional.

One option would be to have this method return an `Optional<Object>` (or `Optional<Identifier>` as per previous comment); alternatively explicitly throw an exception so that the reader knows it's expected.

service/src/main/java/uk/nhs/adaptors/gp2gp/TransformJsonToXml.java
```diff
@@ -0,0 +1,225 @@
...
+        return new String(Files.readAllBytes(Paths.get(file)));
+    }
+
+    private static void writeToFile(String xml, String sourceFileName) {
+        String outputFileName = FilenameUtils.removeExtension(sourceFileName);
+        try {
+            BufferedWriter writer = new BufferedWriter(new FileWriter(XML_OUTPUT_PATH + outputFileName + ".xml"));
+            writer.write(xml);
+            writer.close();
+            LOGGER.info("Contents of file: " + sourceFileName + ". Saved to: " + outputFileName + ".xml");
```
andrewwa-kainos, line 125: Normally considered best practice to use slf4j's log templating:
```suggestion
            LOGGER.info("Contents of file: {}. Saved to: {}.xml", sourceFileName, outputFileName);
```

transformJsonToXml/TransformJsonToXml.sh
```diff
@@ -0,0 +1,3 @@
+cd "../"
+cd "service"
+./gradlew run
```
andrewwa-kainos, line 3: Shell scripts are often `cat`ted etc, so having that new line at the end is useful not to mess with people's terminals.
```suggestion
./gradlew run

```

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredTaskExecutor.java
```diff
@@ -73,4 +73,6 @@ public void execute(GetGpcStructuredTaskDefinition structuredTaskDefinition) {
 
         detectTranslationCompleteService.beginSendingCompleteExtract(ehrExtractStatus);
     }
+
+
```
andrewwa-kainos, line 77: Try to check your changelists before committing / pushing to avoid unnecessary changes

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,237 @@
...
+        LOGGER.error("end");
+    }
+
+    private static InputWrapper getFiles() throws Exception {
+        File[] files = new File(JSON_FILE_INPUT_PATH).listFiles();
+        List<String> jsonStringInputs = new ArrayList<>();
+        List<String> fileNames = new ArrayList<>();
+
+        assert files != null;
+        LOGGER.error(JSON_FILE_INPUT_PATH);
```
msolecka, line 93: The log is displays as error, while it could be valid case as well so I would change it to info and provide something that gives more information.
I would also change the order here, to check files number first and do not use assert (as I don't think is gives us the best information what is the problem)

for example:
```
if (files  == null || files.length == 0) {
    throw new Exception("No json files found");
}
LOGGER.info("Processing " + files.length + " files from location: " + JSON_FILE_INPUT_PATH);
```

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,237 @@
...
+
+        Arrays.stream(files)
+                .peek(file -> LOGGER.info("Parsing file: {}", file.getName()))
+                .filter(file -> FilenameUtils.getExtension(file.getName()).equalsIgnoreCase("json"))
+                .forEach(file -> {
+                    String jsonAsString = null;
+                    try {
+                        jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + file.getName());
+                    } catch (Exception e) {
+                        LOGGER.info("Could not read {}", file.getName());
```
msolecka, line 106: this log would be good to be error, and print exception because it is handled while there was some problem with reading file.

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,237 @@
...
+        messageContext.resetMessageContext();
+
+        return hl7TranslatedResponse;
+    }
+
+    @Data
+    @Builder
+    @AllArgsConstructor
+    public static class InputWrapper {
+        private List<String> jsonFileNames;
```
msolecka, line 234: if InputWrapper would have only 2 strings, jsonInputFile and jsonFileName it would simplify the code much.

So in general you would have 
```
public static class InputFile {
private String jsonFileName;
 private String jsonFileInput;
}
```

then method getFile() would return list of InputFile

```
 private static List<InputFile> getFiles() {
 
 ...
 
 
 return  Arrays.stream(files)
                .peek(file -> LOGGER.info("Parsing file: {}", file.getName()))
                .filter(file -> FilenameUtils.getExtension(file.getName()).equalsIgnoreCase("json"))
                .map(file -> {
                    try {
                        String jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + file.getName());
                        return new InputFile(file.getName(), jsonAsString);
                    } catch (IOException e) {
                       /// throw exception
                    }
                   
                }).collect(Collectors.toList());

```

then in main only:

```
if (startTest.equals("True")) {
                getFiles().forEach(file -> {
                    String xmlResult = mapJsonToXml(file.getJsonFileInput());
                    writeToFile(xmlResult, file.getJsonFileName());
                });
            }
 ```

File: service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java [hunk 29]
```diff
@@ -0,0 +1,232 @@
...
+        LOGGER.error("end");
+    }
+
+    private static InputWrapper getFiles() throws Exception {
+        File[] files = new File(JSON_FILE_INPUT_PATH).listFiles();
+        List<String> jsonStringInputs = new ArrayList<>();
+        List<String> fileNames = new ArrayList<>();
+
+        if (files == null || files.length == 0) {
+            throw new Exception("No json files found");
```
Broom94, line 89: use custom exception for this

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,232 @@
...
+
+        Arrays.stream(files)
+            .peek(file -> LOGGER.info("Parsing file: {}", file.getName()))
+            .filter(file -> FilenameUtils.getExtension(file.getName()).equalsIgnoreCase("json"))
+            .forEach(file -> {
+                String jsonAsString = null;
+                try {
+                    jsonAsString = readJsonFileAsString(JSON_FILE_INPUT_PATH + file.getName());
+                } catch (Exception e) {
+                    LOGGER.error("Could not read {}", file.getName());
```
Broom94, line 102: use custom exception and rethrow

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,232 @@
...
+                }
+                jsonStringInputs.add(jsonAsString);
+                fileNames.add(file.getName());
+            });
+        return InputWrapper.builder().jsonFileInputs(jsonStringInputs).jsonFileNames(fileNames).build();
+    }
+
+    private static String mapJsonToXml(String jsonAsStringInput) throws Exception {
+        final Bundle bundle = new FhirParseService().parseResource(jsonAsStringInput, Bundle.class);
```
Broom94, line 111: try separate class initialisation here so class are loaded once at the start instead of every file

Broom94 (changes requested):
Fix coding style as discussed on teams

service/build.gradle
```diff
@@ -95,6 +95,11 @@ test {
 	useJUnitPlatform()
 }
 
+task(runSimple, dependsOn: 'classes', type: JavaExec) {
```
Broom94, line 98: change name 'runSimple' to something more descriptive

e2e-tests/src/main/java/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,4 @@
+Manifest-Version: 1.0
+Main-Class: uk.nhs.adaptors.gp2gp.transformjsontoxmltool.source.Transfor
+ mJsonToXml
+
```
Broom94, line 4: is this needed?

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/Manifest/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,4 @@
+Manifest-Version: 1.0
+Main-Class: uk.nhs.adaptors.gp2gp.transformjsontoxmltool.source.Transfor
+ mJsonToXml
+
```
Broom94, line 4: is this needed?

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/manifest.txt
```diff
@@ -0,0 +1 @@
+Main-Class: TransformJsonToXml
```
Broom94, line 1: is this needed?

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/source/manifest.txt
```diff
@@ -0,0 +1 @@
+Main-Class: TransformJsonToXml
```
Broom94, line 1: is this needed?

transformJsonToXml/TransformJsonToXml.sh
```diff
@@ -0,0 +1,2 @@
+cd "../service/"
+./gradlew runSimple
```
Broom94, line 2: change name 'runSimple' to something more descriptive

README.md
```diff
@@ -287,6 +287,27 @@ We provide a mock MHS adaptor for local development and testing.
 | MOCK_MHS_SERVER_PORT                 | 8081                      | The port on which the mock MHS Adapter will run.
 | MOCK_MHS_LOGGING_LEVEL               | INFO                      | Mock MHS logging level. One of: DEBUG, INFO, WARN, ERROR. The level DEBUG **MUST NOT** be used when handling live patient data.
 
+## How to transform arbitrary json ASR payload files
+
+This is a document on how to run tests on json files to convert them into the Xml Files for analysis
```
Broom94, line 292: Remove 'This is a document on'

Maybe something like This is a interoperability testing toolt to transform arbitrary/ad-hoc json ASR payloads and access the outputs

README.md
```diff
@@ -287,6 +287,27 @@ We provide a mock MHS adaptor for local development and testing.
...
+
+#### Step 1.
+
+Please locate the file Input folder \
+("integration-adaptor-gp2gp/transformJsonToXml/input/") \
```
Broom94, line 297: follow how this have been done in other areas of readme `integ..`

README.md
```diff
@@ -287,6 +287,27 @@ We provide a mock MHS adaptor for local development and testing.
...
+#### Step 2.
+
+Please locate the TransformJsonToXml.sh file\
+("integration-adaptor-gp2gp/transformJsonToXml/")\
+Run that file
+
+#### Step 3.
+
+The Converted .Xml files will be located in \
+("sintegration-adaptor-gp2gp/transformJsonToXml/output/")
```
Broom94, line 309: no s here

README.md
```diff
@@ -287,6 +287,25 @@ We provide a mock MHS adaptor for local development and testing.
...
+This is an interoperability testing tool to transform arbitrary/ad-hoc json ASR payloads and access the outputs.
+
+1. Navigate to the input folder and place all Json files to convert here.
+`integration-adaptor-gp2gp/transformJsonToXml/input/` 
+   
+
+2. Navigate to the TransformJsonToXml.sh file and run that script to execute the testing tool. 
+   `integration-adaptor-gp2gp/transformJsonToXml/`
+   ```shell script
+   cd ../transformJsonToXml
```
Broom94, line 301: starting from root you wont need to ../

README.md
```diff
@@ -287,6 +287,25 @@ We provide a mock MHS adaptor for local development and testing.
 | MOCK_MHS_SERVER_PORT                 | 8081                      | The port on which the mock MHS Adapter will run.
 | MOCK_MHS_LOGGING_LEVEL               | INFO                      | Mock MHS logging level. One of: DEBUG, INFO, WARN, ERROR. The level DEBUG **MUST NOT** be used when handling live patient data.
 
+### How to transform arbitrary json ASR payload files
+
+This is an interoperability testing tool to transform arbitrary/ad-hoc json ASR payloads and access the outputs.
+
+1. Navigate to the input folder and place all Json files to convert here.
+`integration-adaptor-gp2gp/transformJsonToXml/input/` 
+
```
Broom94, line 296: make shell script for 1 and 3

Broom94 (changes requested):

README.md
Broom94: are you starting from input dir?

Broom94 (approved):

README.md
```diff
@@ -287,6 +287,25 @@ We provide a mock MHS adaptor for local development and testing.
...
+`integration-adaptor-gp2gp/transformJsonToXml/input/` 
+   
+
+2. Navigate to the TransformJsonToXml.sh file and run that script to execute the testing tool. 
+   `integration-adaptor-gp2gp/transformJsonToXml/`
+   ```shell script
+   cd transformJsonToXml
+   ./TransformJsonToXml.sh
+    ```  
+
```
Broom94, line 304: remove extra spaces

Broom94 (approved):

Broom94 (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,250 @@
...
+            for (int i = 0; i < jsonFileInputs.size(); i++) {
+                String jsonString = jsonFileInputs.get(i);
+                String xmlResult = mapJsonToXml(jsonString);
+                String fileName = jsonFileNames.get(i);
+                writeToFile(xmlResult, fileName);
+            }
+        } catch (NHSNumberNotFound | UnreadableJsonFileException | NoJsonFileFound e) {
+            LOGGER.error("error: " + e.getMessage());
+        }
+        LOGGER.error("end");
```
bartek-sarul, line 132: `LOGGER.info`

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,250 @@
...
+import uk.nhs.adaptors.gp2gp.ehr.mapper.RequestStatementMapper;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.StructuredObservationValueMapper;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.SupportedContentTypes;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.diagnosticreport.DiagnosticReportMapper;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.diagnosticreport.ObservationMapper;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.diagnosticreport.SpecimenMapper;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.parameters.EhrExtractTemplateParameters;
+import uk.nhs.adaptors.gp2gp.gpc.GetGpcStructuredTaskDefinition;
+
+public class TransformJsonToXml {
```
bartek-sarul, line 58: If you would make this a console SpringBoot app, you wouldn't have to initialise all those fields you have here. Unless there is some specific thing about this that I don't see

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,250 @@
...
+            .build();
+
+        final EhrExtractTemplateParameters ehrExtractTemplateParameters =
+            EHR_EXTRACT_MAPPER.mapBundleToEhrFhirExtractParams(getGpcStructuredTaskDefinition, bundle);
+
+        final String ehrExtractContent = EHR_EXTRACT_MAPPER.mapEhrExtractToXml(ehrExtractTemplateParameters);
+
+        final String hl7TranslatedResponse = OUTPUT_MESSAGE_WRAPPER_MAPPER.map(getGpcStructuredTaskDefinition, ehrExtractContent);
+
+        MESSAGE_CONTEXT.resetMessageContext();
```
bartek-sarul, line 186: Would be better to wrap code above into a `try-catch` and put the `MESSAGE_CONTEXT.resetMessageContext()` in the `finally` block

service/src/main/java/uk/nhs/adaptors/gp2gp/transformJsonToXmlTool/TransformJsonToXml.java
```diff
@@ -0,0 +1,250 @@
...
+
+        return hl7TranslatedResponse;
+    }
+
+    private static void writeToFile(String xml, String sourceFileName) {
+        String outputFileName = FilenameUtils.removeExtension(sourceFileName);
+        try (BufferedWriter writer =
+                 new BufferedWriter(new FileWriter(XML_OUTPUT_PATH + outputFileName + ".xml", StandardCharsets.UTF_8))) {
```
bartek-sarul, line 194: Check if `FileWriter` doesn't need closing

bartek-sarul (changes requested):

bartek-sarul (approved):

bartek-sarul (approved):

msolecka (approved):

PR #242, opened 2021-05-24
NIAD-1337: Add fromAsid override variable

Broom94:


bartek-sarul (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/request/EhrExtractRequestHandler.java
```diff
@@ -119,14 +126,32 @@ private void createGetGpcStructuredTask(EhrExtractStatus ehrExtractStatus) {
             getRequiredValue(payload, NHS_NUMBER_PATH),
             getRequiredValue(header, FROM_PARTY_ID_PATH),
             getRequiredValue(header, TO_PARTY_ID_PATH),
-            getRequiredValue(payload, FROM_ASID_PATH),
-            getRequiredValue(payload, TO_ASID_PATH),
+            getFromAsid(getRequiredValue(payload, FROM_ASID_PATH)),
```
mattd-kainos, line 129: I don't think we can apply the override here. This will also affect the ASIDs we send back to the requesting system, which if overridden, will be wrong and cause the message to be rejected. We should apply the override when we populate the Ssp headers for the GPC requests.

mattd-kainos (changes requested):
override needs to be applied to headers, not to the entire transfer process

docker/docker-compose.yml
```diff
@@ -35,6 +35,8 @@ services:
       - GP2GP_SSL_TRUST_STORE_URL
       - GP2GP_SSL_TRUST_STORE_PASSWORD
       - GP2GP_GPC_OVERRIDE_NHS_NUMBER
+      - GP2GP_GPC_OVERRIDE_FROM_ASID
+      - GP2GP_GPC_OVERRIDE_TO_ASID
```
mattd-kainos, line 39: Need to update README to include these

mock-mhs-adaptor/src/main/java/uk/nhs/adaptors/mockmhsservice/common/OutboundMessage.java
```diff
@@ -22,6 +22,7 @@
     private String payload;
     @JsonInclude(JsonInclude.Include.NON_NULL)
     private List<Attachment> attachments;
+    @JsonProperty("external_attachments")
```
mattd-kainos, line 25: a mock MHS update is probably needed now?

docker/docker-compose.yml
mattd-kainos: Also consider: we do actually need both of these? We only use `GP2GP_GPC_OVERRIDE_TO_ASID` right?

ElliotB-96 (approved):

mattd-kainos (approved):

docker/docker-compose.yml
mattd-kainos: It seems that SSP in Int is stricter than it used to be. Good to know, we may actually need to use these in production.

PR #244, opened 2021-05-27
NIAD-1464 Ensure standalone Observations are mapped after DiagnosticReports

lgonczar:
Ensure standalone Observations are mapped after DiagnosticReports. Otherwise, there's a possibility of an observation being mapped twice:
- as part of DiagnosticReport processing
- as a standalone Observation

Changes in UAT test files should only be about IDs and the order of standalone components. Their lengths haven't changed.


mattd-kainos (approved):
Looks great!

msolecka (approved):

PR #245, opened 2021-05-27
NIAD-910: Fixes to NIAD-910 after testing.

matthewdoreilly:
Main changes made:

* For an observation, if there are no derived observations from that observation
  * create an Observation Statement (if exists from `prepareObservationStatement` method) and append Narrative Statements (if exists from `prepareNarrativeStatements` method)

* Observation Statement should be output if Compound Statement is a CLUSTER. Observation Statement should not be output if Compound Statement is a BATTERY.

* Specimen Narrative Statement follows same structure as other Narrative Statements.

* When mapping observations (proper observations or dummy observation), if the specimen or observations is dummy, then associate those observations with the specimen


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -120,6 +125,50 @@ public String map() {
             );
         }
 
+        private String prepareObservationStatement(Observation observation) {
```
matthewdoreilly, line 128: No changes here. Just moving these methods further up the class

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -120,7 +127,51 @@ public String map() {
             );
         }
 
-        private String prepareNarrativeStatements(Observation observation) {
+        private Optional<String> prepareObservationStatement(Observation observation, CompoundStatementClassCode classCode) {
```
matthewdoreilly, line 130: GitHub diff may look a bit strange here, only because I've moved some methods up the class for readability. `prepareObservationStatement` was moved up the class, and now returns `<Optional> String` rather than `String`

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -227,50 +278,6 @@ private String prepareStatementsForDerivedObservations(List<Observation> derived
             return derivedObservationsBlock.toString();
         }
 
-        private String prepareObservationStatement(Observation observation) {
```
matthewdoreilly, line 230: See comment I made above about moving methods up the class

andrewwa-kainos (approved):

Broom94 (approved):

PR #253, opened 2021-06-07
NIAD-1464 Use IdMapper when creating NarrativeStatement elements

lgonczar:
Before this change, the fact of mapping an Observation to a NarrativeStatement wasn't registered, which allowed it to be duplicated in the output.

Also, test cases specific to the problem of duplication/omission of observations in mapping have been added.


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -248,6 +250,10 @@ private String mapObservationToNarrativeStatement(Observation observation, Strin
             return TemplateUtils.fillTemplate(NARRATIVE_STATEMENT_TEMPLATE, narrativeStatementTemplateParameters.build());
         }
 
+        private void markObservationAsMapped(Observation observation) {
+            idMapper.getOrNew(ResourceType.Observation, observation.getIdElement());
+        }
```
lgonczar, line 255: This is how we make sure that this observation is marked as mapped, so that it doesn't end up in the output XML as a standalone statement (since it has been mapped to a nested narrative statement).

msolecka: creating dedicated method in IdMapper sounds good to me.

msolecka (approved):

matthewdoreilly (dismissed):

matthewdoreilly:
One of the E2E tests is failing on this branch, `EhrExtractTest > When_ExtractRequestReceived_Expect_ExtractStatusAndDocumentDataAddedToDatabase`

mattd-kainos (changes requested):
I'm concerned this needs more consideration. Are there ever references to the NarrativeStatement with the random id? If so we need to ensure the correct id is referenced.

lgonczar:
> I'm concerned this needs more consideration. Are there ever references to the NarrativeStatement with the random id? If so we need to ensure the correct id is referenced.

I'm not sure what you mean. This method generates a random ID, assigns it to a template and returns a string being an XML fragment. I don't see how any references to those randomly generated IDs could exist, since they don't leave the scope of the method in which they're created (other than being a part of some bigger string).

mattd-kainos:
@dmcavenue1 could you have a look at these changes? We are in effect creating a hidden (unused) HL7 II for the Observation that generates NarrativeStatements in a diagnostic report mapping.
1) If anything else references that Observation then the reference will be invalid (HL7 II was never used in the EHR)
2) The NarrativeStatements have random HL7 II and cannot be referenced from elsewhere in the EHR

dmcavenue1:
> @dmcavenue1 could you have a look at these changes? We are in effect creating a hidden (unused) HL7 II for the Observation that generates NarrativeStatements in a diagnostic report mapping.
> 
> 1. If anything else references that Observation then the reference will be invalid (HL7 II was never used in the EHR)
> 2. The NarrativeStatements have random HL7 II and cannot be referenced from elsewhere in the EHR

Should be fine (subject to trying it) but all association in the GP2GP pathology structure is by containment in the different levels of CompoundStatement so there are no linkages between statements to be preserved at the level of these ObservationStatement or NarrativeStatement. I'm 99% confident that there will be no issues arising. The only thing that is a concern and not so much about DiagnosticReport is the potential for a problem linkage to an Observation to be made in the source system. For that reason it would be better to map UUIDs for Observations even if it meat GUID duplication between CompoundStatement/ObservationSattement where the Observation has been Compounded to accommodate text carrying NarrativeStatement There will be no problem generating a fresh UUID for NarrativeStatement in all cases.

lgonczar:
I propose that we move this discussion outside this PR, as it doesn't change anything with respect to the presented concerns. Its goal is to prevent outputting standalone narrative statements when the source observations have already been mapped as part of diagnostic reports. Other than that, the behaviour stays the same.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
mattd-kainos: This observation is already "mapped" to the CompoundStatement on line 108. Unless I'm missing something this shouldn't be needed.

lgonczar: This happens only when the observation has some derived observations. If it doesn't, then:
- If it's mapped to an ObservationStatement, IdMapper will be used for generating its ID
- If it's mapped to NarrativeStatements, each of those statements will receive a new random ID, but the fact that the Observation has been mapped to something won't be registered (before this PR), as IdMapper won't be used

What we could do is generate the ID using IdMapper the same way it's done in line 108, but move that bit to the top of the method and give the variable a bit more generic name. Then, I can safely remove `markObservationAsMapped`. @mattd-kainos please let me know what you think.

mattd-kainos (commented):
reset review?

msolecka (approved):

matthewdoreilly (approved):

PR #257, opened 2021-06-09
NIAD-846: Send NACK on error

msolecka:

PR #260, opened 2021-06-14
NIAD-1166 Fail the whole conversation when the processing of inbound message or task fails

lgonczar:
Fail the whole conversation when the processing of inbound message or task fails, with the following exceptions:
- message is invalid
- extract doesn't exist in the DB
- send NACK task is being processed

In the above cases, the message won't be acknowledged, so that it can be processed again (or dead-lettered eventually).


service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/ProcessFailureHandlingService.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.nhs.adaptors.gp2gp.ehr.EhrExtractStatusRepository;
+import uk.nhs.adaptors.gp2gp.ehr.EhrExtractStatusService;
+import uk.nhs.adaptors.gp2gp.ehr.SendNegativeAcknowledgementTaskDispatcher;
+
+@Component
+@Slf4j
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class ProcessFailureHandlingService {
```
lgonczar, line 14: Not sure if I should use `conversation` instead of `process`. Thoughts?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/task/TaskHandler.java
```diff
@@ -19,21 +21,49 @@
...
+    public boolean handle(Message message) {
+        TaskDefinition taskDefinition = null;
+
+        try {
+            taskDefinition = readTaskDefinition(message);
+
+            if (!processFailureHandlingService.hasProcessFailed(taskDefinition.getConversationId())) {
+                executeTask(taskDefinition);
+            } else {
+                LOGGER.warn(
```
msolecka, line 40: Shouldn't aborted task return false to not ack the message?

lgonczar: At this point we know the process (conversation?) has failed and we're not going to perform any more tasks. The message can be considered processed, as we've done everything we were supposed to with it.

We should exclude "Send NACK" task from that, though, because NACK message should always be sent - I'll update the code.

msolecka: yes, looking at it form this perspective makes sense, thanks for clarifying.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/ProcessFailureHandlingService.java
msolecka: I would stay with process.

msolecka (commented):
Looks good!
Two questions:
- Are e2e tests going to be in different PR ? it would be good to add it here, we would know already it is working fine
- Modifying data in db with addition information is good, I would only update the story description to avoid misunderstanding while testing.

lgonczar:
> * Are e2e tests going to be in different PR ? it would be good to add it here, we would know already it is working fine

As both this PR and https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/257 cover the same scenario, I thought it would be best to use one e2e test for both and just add an additional assertion for verifying error information in DB.

> * Modifying data in db with addition information is good, I would only update the story description to avoid misunderstanding while testing.

Updating the DB is a part of the description of the story. It mentions information about whether an error occurred plus details, but no specific information on those. I'll add a comment to the story to make it clearer.

msolecka (approved):

msolecka (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -107,9 +117,9 @@ public EhrExtractStatus updateEhrExtractStatusAccessStructured(GetGpcStructuredT
     }
 
     public EhrExtractStatus updateEhrExtractStatusAccessDocument(GetGpcDocumentTaskDefinition documentTaskDefinition,
-            String documentName,
-            String taskId,
-            String messageId) {
+                                                                 String documentName,
+                                                                 String taskId,
+                                                                 String messageId) {
```
Broom94, line 122: I dont believe this is our coding format have you seen this https://gpitbjss.atlassian.net/wiki/spaces/NIA/pages/2108522539/Java+Coding+Standards

intellij formatted at bottom

lgonczar: I guess this came from the fact that the first parameter is now in a new line and IntelliJ wanted to align the rest with this one. I'll change it back.

lgonczar: done

Broom94 (approved):

msolecka (approved):

PR #266, opened 2021-06-21
NIAD-1449 Prevent the execution of JSON to XML transform tool during tests

lgonczar:
* Prevent the execution of JSON to XML transform tool during tests (or app execution)
* Make GIT ignore everything in the tool's output directory, as we don't want to keep those files in the repository.


Broom94:
Deleting the file in the output folder also removes the output folder. This will cause an error for anyone using this tool when the tool tries to save files. The only other thing that could be done if you want to remove is have instruction to create this folder or may a script to?

lgonczar:
> Deleting the file in the output folder also removes the output folder. This will cause an error for anyone using this tool when the tool tries to save files. The only other thing that could be done if you want to remove is have instruction to create this folder or may a script to?

I've put a .gitignore file in that folder. It's not supposed to be removed.

Broom94 (approved):
nice

matthewdoreilly (approved):

msolecka (approved):

PR #268, opened 2021-06-23
NIAD-1285: Add additional test for another gpconnect provider

Broom94:


e2e-tests/src/test/java/uk/nhs/adaptors/gp2gp/e2e/EhrExtractTest.java
```diff
@@ -38,6 +38,7 @@
     private static final String FROM_ASID = "200000000359";
     private static final String TO_ASID = "918999198738";
     private static final String FROM_ODS_CODE = "GPC001";
+    private static final String FROM_ODS_CODE2 = "GP0001";
```
lgonczar, line 41: Worth changing the value to make sure those two don't get mixed somehow.

lgonczar: Sorry, I'm blind, they are different. Maybe `2` instead of `1`, to make the difference more visible? Not so important, up to you.

e2e-tests/src/test/java/uk/nhs/adaptors/gp2gp/e2e/EhrExtractTest.java
```diff
@@ -100,23 +84,55 @@ public void When_ExtractRequestReceivedForPatientWithNoDocs_Expect_DatabaseToBeU
...
 
-    private String buildEhrExtractRequest(String conversationId, String notExistingPatientNhsNumber) throws IOException {
+    @Test
+    public void When_ExtractRequestReceivedDifferentGPCProvider_Expect_ExtractStatusAndDocumentDataAddedToDatabase() throws Exception {
+        String conversationId = UUID.randomUUID().toString();
+        String ehrExtractRequest = buildEhrExtractRequest(conversationId, EXISTING_PATIENT_NHS_NUMBER, FROM_ODS_CODE2);
+        MessageQueue.sendToMhsInboundQueue(ehrExtractRequest);
+
+        assertHappyPathWithDocs(conversationId, FROM_ODS_CODE2);
+    }
```
lgonczar, line 106: This test relies on the existence of another test, which uses a different value for `from ODS code`. In order to make it independent, maybe we should send two messages in it, each with a different ODS code and a different conversation ID and then have two assertions like this:
```
assertHappyPathWithDocs(conversationId1, FROM_ODS_CODE);
assertHappyPathWithDocs(conversationId2, FROM_ODS_CODE2);
```

Broom94: Our first test in e2e sends the same thing but different ods code, however with our await when sending two in one test there is a big delay so I thought this was the best option. What do you think?

lgonczar: If we send both messages before running the assertions, the combined wait time should be similar (but we can check how big the difference is).

Tests need to be independent from each other, so that a change in one test doesn't affect any other. In this case, the removal of the other test would make this one incomplete, but it wouldn't be obvious - the test would still be passing.

msolecka (approved):

lgonczar (approved):

PR #269, opened 2021-06-24
NIAD-1482: Add gpc 1.5.1 to docker-compose

Broom94:


mattd-kainos:
Consider adding a vars.sh example for running the adaptor against the mock?

Broom94:
There are no additional variables. This is only to unblock GPC work for this sprint

matthewdoreilly (commented):
Do we still need the `gpcc` service in docker-compose with this new Docker image being added?

Broom94:
yes we do, the new docker images provided is to replace the wiremock / public demonstrator requests from gpcc

ElliotB-96 (dismissed):
Looks good to me, able to run locally

msolecka (approved):

docker/docker-compose.yml
```diff
@@ -93,3 +93,54 @@ services:
...
+    environment:
+      - TZ=Europe/London
+    links:
+      - gpconnect-api
+    depends_on:
+      - gpconnect-api
+    
+networks: 
+  gpconnect-network-1.5:
+    driver: bridge
```
lgonczar, line 146: Should we use the same network for GPC services as for GP2GP? Not sure if the containers are going to be able to talk to each other.

PR #273, opened 2021-06-28
NIAD-1166: create integration tests for the handling of inbound messages

lgonczar:
These tests are focused on handling inbound messages and their purpose is to verify such things as acknowledging queue message, (the fact of) processing the message and failing the process/conversation.

### DIFF
--- a/service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
+++ b/service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
@@ -1,0 +1,184 @@
+package uk.nhs.adaptors.gp2gp.ehr;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+import static uk.nhs.adaptors.gp2gp.ehr.EhrStatusConstants.CONVERSATION_ID;
+
+import javax.jms.Message;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import lombok.SneakyThrows;
+import uk.nhs.adaptors.gp2gp.common.service.XPathService;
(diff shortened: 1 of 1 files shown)


service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
```diff
@@ -0,0 +1,184 @@
...
+    private EhrExtractStatusRepository ehrExtractStatusRepository;
+    @Mock
+    private Message message;
+    @Mock
+    private InboundMessage inboundMessage;
+
+    @Test
+    @SneakyThrows
+    public void When_MessageIsUnreadable_Expect_MessageProcessingToBeAborted() {
+        mockNonJsonMessage();
```
matthewdoreilly, line 70: This is testing a `JsonProcessingException` is thrown when the inbound queue receives a inbound MHS message, which I don't think would ever happen?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
```diff
@@ -0,0 +1,184 @@
...
+        inboundMessageConsumer.receive(message);
+
+        verify(message, never()).acknowledge();
+        verifyNoInteractions(taskDispatcher);
+    }
+
+    @Test
+    @SneakyThrows
+    public void When_MessageProcessingFails_Expect_WholeProcessToBeFailed() {
+        // ARRANGE
```
matthewdoreilly, line 81: Similarly to your other PR, I'd remove these kind of comments

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
lgonczar: We use ObjectMapper for deserialising inbound messages. If the mapper fails to do that, it throws a `JsonProcessingException` - this is the condition we're creating to verify that the service won't acknowledge the message in such case, and that it won't trigger any task.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
lgonczar: done

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
```diff
@@ -0,0 +1,174 @@
...
+        inboundMessageConsumer.receive(message);
+
+        verify(message, never()).acknowledge();
+        verifyNoInteractions(taskDispatcher);
+    }
+
+    @Test
+    @SneakyThrows
+    public void When_MessageProcessingFails_Expect_WholeProcessToBeFailed() {
+        mockInboundMessage("unsupportedInteractionId", CONTINUE_MESSAGE_PAYLOAD);
```
mattd-kainos, line 81: Ideally, for an invalid interaction id I would expect that we don't attempt to perform any processing on the message at all regardless of ConverationId value.

If we encounter an unknown interaction id that means someone has sent a message to an adaptor which is entirely unrelated to GP2GP.

I think it would be acceptable to raise a bug for this if the fix is difficult or out of scope. Both the impact and likelihood are very low.

msolecka (dismissed):

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
lgonczar: I've changed the cause of failure in this test, so that it's not unsupported interaction ID. When https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/262 is merged, I'll cause message processing to fail in such scenario.

mattd-kainos (approved):

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/InboundMessageHandlingTest.java
lgonczar: I've realised that #262 already handles this case. Just needs to be merged.

msolecka (approved):

PR #276, opened 2021-07-06
NIAD-1476: Observation interpretation.text not being processed.

matthewdoreilly:
`interpretation.text` and `comment` in the Observation resources in FHIR should be concatenated together with “Interpretation: “ prepended (i.e. “Interpretation: {{interpretation.text}} {{comment}})”.

Fix should be made for all occurrences of Observation i.e. when Observations are processed via DiagnosticReport and the more general case of Observation (respecting text handling differences between the two i.e. NarrativeStatement vs pertinentInformation)


service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/DiagnosticReportMapperTest.java
```diff
@@ -1,4 +1,4 @@
-package uk.nhs.adaptors.gp2gp.ehr.mapper;
+package uk.nhs.adaptors.gp2gp.ehr.mapper.diagnosticreport;
```
matthewdoreilly, line 1: Moving `DiagnosticReportMapperTest` to the "diagnosticreport" sub-folder in the unit tests directory.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/SpecimenMapperTest.java
```diff
@@ -1,4 +1,4 @@
-package uk.nhs.adaptors.gp2gp.ehr.mapper;
+package uk.nhs.adaptors.gp2gp.ehr.mapper.diagnosticreport;
```
matthewdoreilly, line 1: Moving `SpecimenMapperTest` to the "diagnosticreport" sub-folder in the unit tests directory.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -161,23 +166,29 @@ private String prepareComment(Observation observation) {
...
-                        commentBuilder.append("Interpretation Text: ").append(interpretation.getText()).append(StringUtils.SPACE);
-                    }
                 });
             }
         }
 
-        commentBuilder.append(observation.hasComment() ? observation.getComment() : StringUtils.EMPTY);
+        Optional<String> interpretationText = StringUtils.isNotBlank(observation.getInterpretation().getText())
+            ? Optional.of(observation.getInterpretation().getText())
+            : Optional.empty();
```
lgonczar, line 183: Just a hint, to apply or not: this condition can be replaced with
```
Optional<String> interpretationText = Optional.ofNullable(observation.getInterpretation().getText())
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -161,23 +166,29 @@ private String prepareComment(Observation observation) {
...
+        String interpretationTextAndComment = Stream.concat(
+            interpretationText.stream(),
+            Optional.ofNullable(observation.getComment()).stream()
```
lgonczar, line 186: Should we do the same with comment as we do with interpretation, i.e. use `StringUtils.isNotBlank`?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -161,23 +166,29 @@ private String prepareComment(Observation observation) {
...
+        ).collect(Collectors.joining(StringUtils.SPACE));
```
lgonczar, line 187: Joining the two parts with a space is what the story requires. I'm wondering, though, if we could replace it with some other character, e.g. `;`? Otherwise, it might be hard to see where one part ends and the other one starts.

matthewdoreilly: @dmcavenue1 What do you think about this?
Making this `Interpretation: {{interpretation.text}}; {{comment}}` rather than `Interpretation: {{interpretation.text}} {{comment}}`?

dmcavenue1: I'm not for it - think we already may have too many different delimiters/separators in text so I'm happier with just the space and ; has a source code feel to me

lgonczar: Not using any kind of separator we're merging two blocks of tests together without providing any information about where one ends and another one starts. That may be confusing to the person who reads it.

As for the change, I'm fine with it (the space was already present in the code), but just pointing at a potential issue. We can discuss offline and treat it unrelated to this PR.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -161,23 +166,26 @@ private String prepareComment(Observation observation) {
...
-                    if (interpretation.hasText()) {
-                        commentBuilder.append("Interpretation Text: ").append(interpretation.getText()).append(StringUtils.SPACE);
-                    }
                 });
             }
         }
 
-        commentBuilder.append(observation.hasComment() ? observation.getComment() : StringUtils.EMPTY);
+        String interpretationTextAndComment = Stream.concat(
+            Optional.ofNullable(observation.getInterpretation().getText()).stream(),
```
lgonczar, line 182: Are we sure that interpretation is not null at this point? We're outside the `if (observation.hasInterpretation()) {...}` clause.

service/src/test/resources/ehr/mapper/diagnosticreport/observation/observation_compound_statement_with_data_absent_reason_and_interpretation_and_body_site_and_method.xml
```diff
@@ -26,23 +26,6 @@
     </ObservationStatement>
 </component>
             <component typeCode="COMP" contextConductionInd="true">
-    <NarrativeStatement classCode="OBS" moodCode="EVN">
```
lgonczar, line 29: Is this NarrativeStatement supposed to disappear?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
matthewdoreilly: Interestingly it doesn't fail. If there is no interpretation in FHIR, `observation.getInterpretation()` will return an object (not "null"), and the properties within the object will be null, so it won't throw a null exception

service/src/test/resources/ehr/mapper/diagnosticreport/observation/observation_compound_statement_with_data_absent_reason_and_interpretation_and_body_site_and_method.xml
matthewdoreilly: Yep, this is the `observation.comment`. You'll see as part of this story this is now appended to the `observation.interpretation.text` in the "AGGREGATE COMMENT SET" NarrativeStatement below, so we don't need it output here (duplicating information)

lgonczar (dismissed):

ElliotB-96 (dismissed):

lgonczar (approved):

ElliotB-96 (approved):

PR #280, opened 2021-07-09
NIAD-1169: Mapping change Observation .valueSampleData .valueAttachment

Broom94:
fail process when attachment or sampleddata is populated, add additional tests for error thrown, remove now unused output file and add additional input file for attachment


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -59,8 +59,11 @@ public String mapObservationToObservationStatement(Observation observation, bool
             Type value = observation.getValue();
 
             if (UNHANDLED_TYPES.contains(value.getClass())) {
-                LOGGER.info("Observation value type {} not supported. Mapping for this field is skipped",
+                LOGGER.info("Observation value type {} not supported. Mapping for transaction is terminated",
                     observation.getValue().getClass());
+                throw new EhrMapperException(
+                    String.format("Observation value type %s not supported. Mapping for transaction is terminated",
+                        observation.getValue().getClass()));
```
lgonczar, line 66: From the mapper's point of view, we've got an error here - unsupported observation value type. I think it's best to remove the part about terminating - this is handled outside the scope of this mapper and it shouldn't care/know what happens next.

Since we're throwing an exception, `LOGGER.info(...)` is unnecessary, as the exception details will be logged, anyway.

lgonczar (dismissed):

sylwestergryzio (approved):

sylwestergryzio (approved):

lgonczar (approved):

PR #285, opened 2021-07-19
Niad 1173 niad 1577 related problem and notes added to observation stmt text

sylwestergryzio:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
```diff
@@ -48,6 +49,12 @@ public String mapCodeableConceptToCd(CodeableConcept codeableConcept) {
         return TemplateUtils.fillTemplate(CODEABLE_CONCEPT_CD_TEMPLATE, builder.build());
     }
 
+    public String getDisplayFromCodeableConcept(CodeableConcept codeableConcept) {
+        return findMainCode(codeableConcept)
+            .map(cc -> findDisplayText(cc).orElse(StringUtils.EMPTY))
+            .orElse(StringUtils.EMPTY);
```
ElliotB-96, line 55: Do you need this second orElse?

ElliotB-96 (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
sylwestergryzio: as discussed, that orElse is in case findMainCode returns empty.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/wrapper/ConditionWrapper.java
```diff
@@ -0,0 +1,121 @@
...
+import uk.nhs.adaptors.gp2gp.ehr.mapper.CodeableConceptCdMapper;
+import uk.nhs.adaptors.gp2gp.ehr.mapper.MessageContext;
+
+@Slf4j
+public class ConditionWrapper {
+
+    private static final String PROBLEM_INFO_PREFIX = "Problem Info: ";
+    private static final String PROBLEM_NOTES_PREFIX = "Problem Notes: ";
+    private static final String NOTES_DELIMITER = "; ";
+    private static Map<String, String> reversedRelMapping =
```
Broom94, line 26: Optional could use unmodifiable or immutable map here https://www.baeldung.com/java-immutable-maps

Broom94 (approved):
Looks good, approving with one optional comment

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/wrapper/ConditionWrapper.java
sylwestergryzio: Done. It would be even more helpful when that variable was exposed, however here we use only internally in a class. But applied the change as good practice.

Broom94 (approved):

ElliotB-96 (approved):

PR #295, opened 2021-08-13
NIAD-1583: Unrecognised encounter codes should display the original term text

rorysproule1:
- In the case of unrecognised encounter code, the original text should first be provided by the .text value (If it is present) otherwise the .display value should be used (always present.)


bbell-kainos (dismissed):

ElliotB-96 (commented):
Could you please fix the formatting so the actual changes are more visible please?

ElliotB-96 (approved):
Looks good

bbell-kainos (approved):

PR #298, opened 2021-08-17
NIAD-1576: availabilityTime of condition should use onSetDateTime or else output a nullFlavor of UNK

rorysproule1:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapper.java
```diff
@@ -225,4 +226,11 @@ private boolean isSuppressedMedicationRequest(Reference reference) {
         // for all other types do not suppress with this function
         return false;
     }
+
+    private Optional<String> buildObservationStatementAvailabilityTime(Condition condition) {
+        if (condition.hasOnsetDateTimeType()) {
```
bartek-sarul, line 231: Can be replaced with a single-liner:
```
return Optional.of(condition)
    .filter(Condition:: hasOnsetDateTimeType)
    .map(Condition:: getOnsetDateTimeType)
    .map(DateFormatUtil::toHl7Format);
```

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapperTest.java
```diff
@@ -49,6 +49,8 @@
 
     private static final String INPUT_JSON_WITH_ACTUAL_PROBLEM_OBSERVATION = CONDITION_FILE_LOCATIONS + "condition_all_included.json";
     private static final String INPUT_JSON_NO_ACTUAL_PROBLEM = CONDITION_FILE_LOCATIONS + "condition_no_problem.json";
+    private static final String INPUT_JSON_NO_ACTUAL_PROBLEM_NO_ONSETDATE = CONDITION_FILE_LOCATIONS + "condition_no_problem_no_onsetdate"
+        + ".json";
```
bartek-sarul, line 53: Why this split?

bartek-sarul (changes requested):

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ConditionLinkSetMapperTest.java
rorysproule1: Must've been an auto reformat, will fix and check to see if it passes the style checks.

bbell-kainos (approved):

PR #322, opened 2021-09-06
NIAD-1603: AllergyIntolerance: Handle asserter as relatedPerson

bbell-kainos:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AllergyStructureMapper.java
```diff
@@ -193,6 +195,16 @@ private String buildAsserterPertinentInformation(AllergyIntolerance allergyIntol
             IIdType reference = allergyIntolerance.getAsserter().getReferenceElement();
             if (reference.getResourceType().equals(ResourceType.Patient.name())) {
                 return PATIENT_ASSERTER;
+            } else if (reference.getResourceType().equals(ResourceType.RelatedPerson.name())) {
+                RelatedPerson relatedPerson = messageContext
+                    .getInputBundleHolder()
+                    .getResource(reference)
+                    .map(RelatedPerson.class::cast)
+                    .get();
```
bartek-sarul, line 203: What if there is no such resource?
It should be something like this:
```
return messageContext
    .getInputBundleHolder()
    .getResource(reference)
    .map(RelatedPerson.class::cast)
    .filter(RelatedPerson::hasName)
    .map(RelatedPerson::getName)
    .map(names -> names.get(0))
    .map(HumanName::GetNameAsSingleString)
    .map(name -> String.format(RELATED_PERSON_ASSERTER, name))
    .orElse(StringUtils.EMPTY);
```

bartek-sarul (changes requested):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AllergyStructureMapper.java
bbell-kainos: Addressed

bartek-sarul (approved):

ElliotB-96 (commented):
Can you include a test where the relatedPerson does not have a name

bartek-sarul (approved):

bbell-kainos:
> Can you include a test where the relatedPerson does not have a name

ElliotB-96 (approved):

PR #334, opened 2021-09-13
Testing SSP Proxy against wiremock

ElliotB-96:
- Added docker/NGINX reverse proxy for testing GPCC SSP against our wiremock


bartek-sarul (approved):

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,21 @@
...
+
+
+server {
+    listen 80;
+
+    access_log /var/log/nginx/access.log;
+    error_log /var/log/nginx/error.log;
+
+    location / {
+         rewrite /http://localhost:8110(.*) /$1  break;
```
sylwestergryzio, line 13: I',m not sure if you need the rewrite rule. It seems like you would like to pass all the parameters as they are, so proxy_pass should be enough. Please check that

sylwestergryzio (approved):

bbell-kainos (approved):

sylwestergryzio (approved):

PR #371, opened 2021-10-05
ReferralRequest: Annotation handling - suppress @ when no author/time 

AlanSimpson1723:


rorysproule1 (dismissed):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/RequestStatementMapper.java
```diff
@@ -341,6 +343,17 @@ private String buildPriorityDescription() {
 
             return StringUtils.EMPTY;
         }
+
+        private String buildConditionalNoteDesc(Annotation value) {
+            String author = extractAuthor(messageContext, value);
+            String time = extractNoteTime(value);
+
+            if (author.equals(StringUtils.EMPTY) && time.equals(StringUtils.EMPTY)) {
```
pawel-wlodarczyk-kainos, line 351: There is no need to compare those strings this way. You could write it this way (my recommendation):
```java
return author.isEmpty() && time.isEmpty()
	? String.format(NOTE_WITH_NO_AUTHOR_AND_TIME, value.getText())
	: String.format(NOTE, author, time, value.getText());
```
or with `if`s:
```java
if (author.isEmpty() && time.isEmpty()) {
	return String.format(NOTE_WITH_NO_AUTHOR_AND_TIME, value.getText());
} else {
	return String.format(NOTE, author, time, value.getText());
}
```

pawel-wlodarczyk-kainos (changes requested):

pawel-wlodarczyk-kainos (approved):

PR #402, opened 2021-10-20
NIAD-1549: DiagnosticReport: Observation.valueString is Observation text and needs to be handled as NarrativeStatement

bbell-kainos:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -240,6 +241,10 @@ private String mapObservationToObservationStatement(MultiStatementObservationHol
             interpretationTextAndComment.append(StringUtils.LF).append(observation.getComment());
         }
 
+        if (observation.hasValueStringType()) {
```
bartek-sarul, line 244: No need to check for `holder.getObservation().hasValue()` the same as it is few lines above?
https://github.com/nhsconnect/integration-adaptor-gp2gp/pull/402/files#diff-60fce59b547f21c0c6a2664153b1a9f2d3e2e3432d3eb8131b10ae658ed43fe6R187

PR #431, opened 2021-11-17
Niad 1757 allergy intollerance change approach to pass drug allergies and dont force degrade

AlanSimpson1723:
Updating PR that was already merged in


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
```diff
@@ -77,12 +77,16 @@ public String mapCodeableConceptToCdForTransformedActualProblemHeader(CodeableCo
                 return Optional.of(codeableConcept.getText());
             } else {
                 var extension = retrieveDisplayExtension(coding.get());
-                return extension.map(value -> value
-                    .getExtension()
-                    .stream()
+                Optional<String> originalText = extension.stream()
                     .filter(extension1 -> extension1.getUrl().equals(DESCRIPTION_DISPLAY))
```
bchotkowski, line 81: Not sure if this shouldn't be following the principle of checking if CONST equals something in order to handle the NullpointerException case?
```suggestion
                    .filter(extension1 -> DESCRIPTION_DISPLAY.equals(extension1.getUrl()))
```

rorysproule1, line 81: Just a small one, might not have even been your change. 

rename extension1 to extension, if needs be then renamed the current extension variable to displayExtension

rorysproule1 (approved):
Approved, just with those 2 small fixes suggested

rorysproule1 (approved):

bchotkowski (approved):

PR #441, opened 2021-11-24
NIAD-1757: AllergyIntolerance change approach to pass drug allergies and dont force degrade

AlanSimpson1723:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/AllergyStructureMapper.java
```diff
@@ -105,9 +105,11 @@ private String buildCode(AllergyIntolerance allergyIntolerance) {
                         .orElse(StringUtils.EMPTY);
 
                     if (category.equals(ENVIRONMENT_CATEGORY)) {
```
rorysproule1, line 107: As you are already filtering for these two category types and they seem to be calling the same function. If they are in fact meant to do the same thing then possibly change the following set of ifs to:

```suggestion
                    if (!category.isEmpty()) {
                        return codeableConceptCdMapper.mapCodeableConceptToCdForAllergy(allergyIntolerance.getCode(),
                                allergyIntolerance.getClinicalStatus());
                    } else {
                        throw new EhrMapperException("Category could not be mapped");
                    }
```

rorysproule1 (approved):

ElliotB-96 (approved):

PR #473, opened 2022-07-28
Niad 2293 ack bugfix

benhession:
### Issue

When an Acknowledge message is received for example from the first part of a message, the EHR, processing stops on any further received messages against the Conversation. 

Therefore, no COPC messages are being processed.

### Overview of changes

- When the structured record is mapped the message ID of the generated EHR Extract is saved to the database.
- When an ACK is received the conversation is only closed if the message reference matches the saved EHR Extract message ID


dfk97 (approved):
All good

PR #474, opened 2022-08-01
Niad 2294 nack does not end processing bug

benhession:
If a NACK message is received that indicates a critical item has failed e.g. the processing of the EHR message, later messages continue to be processed.

### Changes made
- Close the conversation if a NACK is received that references the EHR Extract
- Accept EHR Extract NACKs while waiting for a continue message.


dfk97 (approved):
Seems good to me

PR #485, opened 2022-10-21
Niad 2395 mapping bug unit of measurements

benhession:
“The unit of measurement is converted to "null" for
        (a) Standing height 186cm
        (b) Blood pressure 129/87 mmHg
        (c) Smoking free weeks: 12 weeks“

Comparing the attached GPC output and GP2GP output, no value units have been mapped in the GP2GP HL7 payload.

Examples:

        (a) Standing height 186cm
        (b) Blood pressure 129/87 mmHg
        (c) Smoking free weeks: 12 weeks

GP2GP Output:

      <value xsi:type="PQ" value="186" unit="null"/>

Action: Fix unit to include the given value in from the GPC output.

**NOTE: The unit of time for `MedicationRequest.dispenceRequest.expectedSupplyDuration`  was also being mapped to `null` in the Medicus example. So I've addressed it as part of this ticket. **


dfk97 (approved):
All good to me

PR #491, opened 2022-11-15
NIAD-2402: Add fromAsid and toAsid to EHR status endpoint.

benhession:
Since the introduction on the ASID override variables in the GP2GP adaptor, it is possible to lose the original ASID provided by the given EHR message.

Scenerio 1

GIVEN an EHR request is received
THEN the toASID and fromASID fields should be recorded against the EHR in the Database. 

Scenerio 2

Given a request is made to the status endpoint for a given Conversation ID
THEN the toAsid and fromAsid fields original given for the EHR request should be returned in the json e.g.

{
    fromASID: “…“,
    toASID: “….“
    ….
}


dfk97 (approved):
All good to me :)

dfk97 (approved):

dfk97 (approved):

dfk97 (approved):

PR #495, opened 2023-01-11
NIAD-2414: Mapping bug - Topic not mapped

benhession:
**Problem**

Topic (areas of discussion within a Consultation) do not appear to be included in the extract message produced by the GP2GP adaptor.

**Work undertaken**

- Mapped GP Connect Topics to GP2GP Compound statements with  `classCode=TOPIC`
- Mapped GP Connect Categories (Headers) to GP2GP Compound statements with `classCode=CATEGORY`
- If a topic has a related problem the snomed code of the problem (if present) is used to code the topic. Otherwise, the snomed code for Unspecified Problem is used. In both cases, the topic's title (if present) is placed in the `originalText` element.


SRAlexander (approved):

SRAlexander:
There's a lot here but the test modifications look suitable for the new topics mappings.

PR #496, opened 2023-02-06
NIAD-2580 - Run1 Cycle1 - Body Site value under Procedure is not displayed 

MartinWheelerMT:
Development Task: When we receive an entry with a Bodysite parameter, we should map the display text of the bodysite to the text/comment section of the parent entry prefixed with “BodySite:”, translating this across into the HL7 format. 

Update Observation statement mapper to map body site correctly 
Add example json to WireMock
Add example test files for parameterised test
Add E2E test for Body Site Mapping


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationStatementMapper.java
```diff
@@ -68,6 +70,11 @@ public String mapObservationToObservationStatement(Observation observation, bool
             }
         }
 
+        var bodySite = extractTextOrCoding(observation.getBodySite())
```
SRAlexander, line 73: IS it worth extracting this into a small sub function if the code is the same in both cases here?

SRAlexander (approved):
Small possible improvement but not something that would block this going through.

PR #499, opened 2023-02-07
task_2579 > main

SRAlexander:
core changes to introduce the medication specific coding mapper and unit test fixes

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
@@ -3,7 +3,6 @@ package uk.nhs.adaptors.gp2gp.ehr.mapper;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
-import java.util.logging.Logger;
 
 import org.apache.commons.lang3.StringUtils;
 import org.hl7.fhir.dstu3.model.AllergyIntolerance;
@@ -95,7 +94,6 @@ public class CodeableConceptCdMapper {
         var builder = CodeableConceptCdTemplateParameters.builder();
         var mainCode = findMainCode(codeableConcept);
 
-        LOGGER.debug("Maincode found? :" + mainCode.isPresent());
         builder.nullFlavor(mainCode.isEmpty());
 
         if (mainCode.isPresent()) {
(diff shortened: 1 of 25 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
```diff
@@ -87,6 +87,43 @@ public String mapCodeableConceptToCd(CodeableConcept codeableConcept) {
...
+        builder.nullFlavor(mainCode.isEmpty());
+
+        if (mainCode.isPresent()) {
+            var extension = retrieveDescriptionExtension(mainCode.get())
+                .map(Extension::getExtension)
+                .orElse(Collections.emptyList());
+
+            builder.mainCodeSystem(SNOMED_SYSTEM_CODE);
+
+            Optional<String> code = Optional.of(mainCode.get().getCode().toString());
```
benhession, line 106: Is it possible for `getCode` to be null here. If so should we use `Optional.ofNullable()` in case there is a `NullPointerException`?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
```diff
@@ -87,6 +87,43 @@ public String mapCodeableConceptToCd(CodeableConcept codeableConcept) {
...
+            code.ifPresent(builder::mainCode);
+
+            Optional<String> displayName = extension.stream()
+                .filter(displayExtension -> DESCRIPTION_DISPLAY.equals(displayExtension.getUrl()))
+                .map(description -> description.getValue().toString())
+                .findFirst()
+                .or(() -> Optional.of(mainCode.get().getDisplay()));
```
benhession, line 113: Same here, could `getDisplay()` return null?  `Optional.of` can throw a `NullPointerException`

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
SRAlexander: Good Shouts, there seems to be a pattern in this class where potential nullables are being ignored. I've now been through the class and corrected all instances.

benhession (approved):

PR #502, opened 2023-07-11
NIAD-2726: BIG_FIX external attachments limit

benhession:
## Description
The MHS adaptor can limit the amount of external attachments. If the maximum value is hit the outbound service will respond with a http status of 400 and a plain text body describing the validation errors.   

According to the document *20151012_GP2GP 2.2 Spec clarifications.doc *:

>Where a sending system is unable to process an extract due to size of the record it must respond to the EHR_Request with an Application Acknowledgement with Response Code 30.

The adaptor is currently responding with response code 99.

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/WebClientFilterService.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/WebClientFilterService.java
@@ -75,7 +75,7 @@ public class WebClientFilterService {
             if (requestType.equals(RequestType.MHS_OUTBOUND) && clientResponse.statusCode().is5xxServerError()) {
                 return Mono.error(new MhsServerErrorException("MHS responded with status code " + clientResponse.statusCode().value()));
             }
-            if (requestType.equals(RequestType.MHS_OUTBOUND) && clientResponse.statusCode().value() == 400) {
+            if (requestType.equals(RequestType.MHS_OUTBOUND) && clientResponse.statusCode().value() == BAD_REQUEST.value()) {
                 return clientResponse.bodyToMono(String.class)
                     .flatMap(WebClientFilterService::handle400FromMhsOutbound);
             }
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/WebClientFilterService.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/WebClientFilterService.java
@@ -1,5 +1,11 @@
 package uk.nhs.adaptors.gp2gp.common.service;
 
+import static org.springframework.http.HttpStatus.BAD_REQUEST;
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
+import static org.springframework.http.HttpStatus.NOT_FOUND;
+import static org.springframework.http.HttpStatus.UNAUTHORIZED;
+import static org.springframework.http.HttpStatus.UNPROCESSABLE_ENTITY;
+
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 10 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/WebClientFilterService.java
```diff
@@ -75,11 +80,29 @@ private static ExchangeFilterFunction errorHandling(RequestType requestType, Htt
...
+                    .flatMap(WebClientFilterService::handle400FromMhsOutbound);
+            }
 
             return getResponseError(clientResponse, requestType);
         });
     }
 
+    private static Mono<ClientResponse> handle400FromMhsOutbound(String body) {
+
+        Pattern pattern = Pattern.compile(MAX_ATTACHMENTS_REGEX, CASE_INSENSITIVE);
```
adrianclay, line 94: This way of checking what the response is feels super janky.

I'm not saying we should do this, but in the ideal world some sort of machine readable API response would be much more preferable from MHS.  I'm surprised that the MHS doesn't already provide this 😭

service/src/main/java/uk/nhs/adaptors/gp2gp/common/task/TaskErrorHandler.java
```diff
@@ -24,45 +50,53 @@ public boolean handleRequestError(TaskDefinition taskDefinition) {
...
                 "06",
                 "Patient not at surgery."
         );
     }
 
+    private boolean handleMaximumExternalAttachmentsError(TaskDefinition taskDefinition) {
+        return handleFailingProcess(
+            taskDefinition,
+            "30",
+            "Large Message general failure"
```
adrianclay, line 96: I'm not sure how this message is used, but it currently feels quite vague.

If we know this is specifically about "too many attachments for the MHS adaptor", including that in the string might be useful, depending on how this string is used.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/task/TaskHandlerTest.java
```diff
@@ -306,15 +320,42 @@ public void When_Handle_WithDataAccessResourceFailureException_Expect_ExceptionT
...
         assertThat(result).isFalse();
 
-        verify(processingErrorHandler).handleGeneralProcessingError(any());
+        verify(taskErrorHandler).handleProcessingError(eq(exception), any());
 
     }
 
+    @Test
+    @SneakyThrows
+    public void When_Handle_WithGpConnectInvalidException_Expect_ProcessFailed() {
```
adrianclay, line 335: I have mixed feelings about being specific about `GpConnectInvalidException` within this unit test.  Feels slightly unnecessary (and potentially misleading), given that there is no behaviour for it.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/WebClientFilterService.java
benhession: Yeah, it would be better if the MHS adaptor had returned the validation errors as JSON. I was surprised it plain text too. I thought this was the best way to check it, given that it currently is.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/task/TaskErrorHandler.java
benhession: Agreed, I took the text from [this document](file:///C:/Users/BenjaminHession/Downloads/NPFIT-PC-BLD-0083.08%20GP2GP%20Response%20Codes%20(1).pdf), but we are using it in a different context here. I'll update it. The values are inserted into the Negative Acknowledgement, however the string is optional. So I think we should be ok to change it. The code is the more important element for the receiving system.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/task/TaskHandlerTest.java
benhession: Yeah, that makes sense. There are several similar tests that can be reduced to one

adrianclay (approved):

MartinWheelerMT (approved):

PR #511, opened 2023-08-18
StructuredRecordMappingService populates a contentType field

adrianclay:
This will allow us to store that field within the database, and use its value when chunking large attachments into chunks.


benhession (commented):
Adding the content field makes sense and I don't believe it should cause any issues. If it's going to be added to the db it should be added to GpcDocument too. Is this a PR a work in progress or does it need approval?

benhession (approved):

PR #513, opened 2023-08-21
Fix duplicate entries within DocumentReferences array

adrianclay:
When checking for Attachments without a URL, the `GetGpcStructuredTaskExecutor` would insert two entries into the Documents array to be stored within the database.

Adds test to cover this behaviour, and removes duplicate code which generates the additional entry.

### DIFF
--- a/service/src/test/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredTaskExecutorTest.java
+++ b/service/src/test/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredTaskExecutorTest.java
@@ -4,6 +4,9 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.NullSource;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import uk.nhs.adaptors.gp2gp.common.configuration.Gp2gpConfiguration;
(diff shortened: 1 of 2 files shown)


service/src/test/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredTaskExecutorTest.java
```diff
@@ -86,6 +89,54 @@ public void When_BundleContainsExternalDocumentReference_Expect_DocumentAddedToE
         );
     }
 
+    @ParameterizedTest
+    @EmptySource
```
adrianclay, line 93: Apparently this `EmptySource` is how you can add NULL to a parameterized test @MartinWheelerMT .

MartinWheelerMT: Good to know, I thought it would be possible

MartinWheelerMT: Actually I think this is wrong, it looks like @NullSource is for nulls, @EmptySource for empty strings, @NullAndEmptySource for both:

@EmptySource is an [ArgumentsSource](https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/params/provider/ArgumentsSource.html) which provides a single empty argument to the annotated @ParameterizedTest method.

@NullSource is an [ArgumentsSource](https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/params/provider/ArgumentsSource.html) which provides a single null argument to the annotated @ParameterizedTest method.

Or combine the two:
@NullAndEmptySource is a composed annotation that combines the functionality of [@NullSource](https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/params/provider/NullSource.html) and [@EmptySource](https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/params/provider/EmptySource.html).

MartinWheelerMT (changes requested):

service/src/test/java/uk/nhs/adaptors/gp2gp/gpc/GetGpcStructuredTaskExecutorTest.java
MartinWheelerMT: For reason it lowercases `@NullSource`

MartinWheelerMT (approved):

PR #523, opened 2023-09-14
NIAD-2814: Fixed plain text file attachment being transferred into TPP as Base64Encoded

ole4ryb:
## What

In the structured record mapping service, we have changed the outbound message which was always being created with a hardcoded value for `originalBase64` as `true`.  This flag appears to be ignored by EMIS, however it affected the import into TPP.

The following changes have been made:

* `originalBase64` will now be hardcoded to `false` when creating the outbound message in the structured record mapping service
* Unit Tests have been updated to reflect this change
* WireMock updated to assist in testing, with new responses created for this

## Why

When transferring a plain text file to TPP the text attachment was stored with the content encoded as Base64 even though the original attachment was not encoded.   In EMIS this attachment is displayed in plain text as expected.  This is not correct behaviour and this could cause confusion when the attachment is viewed.

This was reported in [NIAD-2814](https://gpitbjss.atlassian.net/browse/NIAD-2814)

## Type of change

## Checklist:

### DIFF
--- a/docker/wiremock/stubs/__files/correctPatientStructuredRecordWithTextDocAttachment.json
+++ b/docker/wiremock/stubs/__files/correctPatientStructuredRecordWithTextDocAttachment.json
@@ -81,19 +81,19 @@
               }
             ],
             "system": "https://fhir.nhs.uk/Id/nhs-number",
-            "value": "9729649553"
+            "value": "9729649561"
           }
         ],
         "active": true,
         "name": [
           {
             "use": "official",
-            "family": "BOWDEN",
+            "family": "CARMEL",
             "prefix": [
               "MS"
             ],
             "given": [
-              "Flora"
+              "Evelyn"
             ]
           }
         ],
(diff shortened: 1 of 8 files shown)


docker/wiremock/stubs/__files/correctPatientStructuredRecordWithTextDocAttachment.json
```diff
@@ -0,0 +1,1209 @@
+{
```
adrianclay, line 1: Were these stubs used for testing within the PTL environment?

docker/wiremock/stubs/__files/correctSimpleTextDocAttachment.json
```diff
@@ -0,0 +1,6 @@
+{
+  "resourceType": "Binary",
+  "id": "care-record-document--30e49be4-42aa-4dff-87d0-3c1c2cc22a6f",
+  "contentType": "text/plain",
+  "content": "SG93IGFyIHlvdQ=="
```
adrianclay, line 5: This content type looks to be Base64.

I don't think in the scenario the Document was encoded in Base64.

Might be worth having a quick call to understand what/how you tested this scenario?!

docker/wiremock/stubs/__files/correctPatientStructuredRecordWithTextDocAttachment.json
ole4ryb: yes

docker/wiremock/stubs/__files/correctSimpleTextDocAttachment.json
ole4ryb: As per the specification retrieved attachment documents are expected to be encoded with binary Base64. 
https://developer.nhs.uk/apis/gpconnect-1-6-0/access_documents_development_binary.html

PR #526, opened 2023-09-21
Document `/requests` method

adrianclay:
## What

Add documentation for the `/requests` endpoint.

## Why

To help our users.

## Type of change

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,72 @@ The response will contain the following fields:
 ```
 </details>
 
+## How to query the Requests endpoint
+
+An API is provided to query the status of all transfers to an incumbent.
+Any "in-progress" transfers are excluded from this list, but become available once they either succeed or fail.
+
+Requests can be made to the following endpoint, where each attribute within the JSON POST body is an optional filter criteria.
+
+```http request
+POST /requests
+Content-Type: application/json
+
+{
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -395,6 +395,72 @@ The response will contain the following fields:
...
+| toAsid                   |                                                                                                                 | string        |                                                                                               | False    |
+| fromOdsCode              |                                                                                                                 | string        |                                                                                               | False    |
+| toOdsCode                |                                                                                                                 | string        |                                                                                               | False    |
+| migrationStatus          | The current state of the transfer, a status of COMPLETE_WITH_ISSUES is given if placeholder documents were sent | string / enum | COMPLETE <br/> COMPLETE_WITH_ISSUES <br/> FAILED_NME <br/> FAILED_INCUMBENT <br/> IN_PROGRESS | False    |
```
benhession, line 433: Can this field be in-progress,  line 401 says they are excluded?

benhession: I guess it's part of the business requirements in-progress transfers are excluded. I would have thought it would be useful to see them.

adrianclay: Assuming that all transfers eventually finish, I can see the logic behind only including "not in-progress" transfers, as that way the values displayed are final.

Whether that logic is right or wrong, I can't say 😄

benhession (approved):

PR #536, opened 2023-11-24
Java heap memory increase to improve overall performance of different…

ole4ryb:
JVM needs sufficient memory in order to run well. 

## What

Java heap memory will be increased for different project components to run faster. It includes our Jenkins integration testing which will benefit from bigger memory allocation..

## Why

## Type of change

## Checklist:


docker/service/Dockerfile
```diff
@@ -20,5 +20,5 @@ COPY --from=package /home/gradle/service/build/libs/gp2gp.jar /app/gp2gp.jar
 
 USER 65534
 
-ENTRYPOINT ["java", "-cp", "/app/gp2gp.jar", "-Dloader.main=uk.nhs.adaptors.gp2gp.Gp2gpApplication", "org.springframework.boot.loader.PropertiesLauncher"]
+ENTRYPOINT ["java", "-XX:+UseG1GC", "-cp", "/app/gp2gp.jar", "-Dloader.main=uk.nhs.adaptors.gp2gp.Gp2gpApplication", "org.springframework.boot.loader.PropertiesLauncher"]
```
adrianclay, line 23: Hmm, [from my reading](https://learn.microsoft.com/en-us/java/openjdk/reasons-to-move-to-java-11#garbage-collection-10) looks like this is unnecessary?

ole4ryb: by setting up this clearly we want to make sure that we are using G1, which is a better GC compared to the other ones. 
Sometimes JVM opt in for a different GC, like a Serial GC etc

ole4ryb: G1 is recommended for application with large datasets. It has become the standard GC in recent years for Java due to its excellent performance. It should be enabled in GP2GP by default (as long as there are more than 1 vCPU) I just want to reinforce that by providing the flag. Here is some brief information about https://en.wikipedia.org/wiki/Garbage-first_collector from wiki.

ole4ryb: @adrianclay I've decided to remove G1GC flag for now since the memory issues for NIAD-2348 was fixed by adjusting the config but still I will be monitoring the performance aspect of GP2GP to see if that flag should be re-introduced in the future.

CHANGELOG.md
```diff
@@ -6,12 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed 
+### Added
+- Added more memory allocation to E2E tests
```
adrianclay, line 10: I don't think end users will care about this?

ole4ryb: I will remove it.

PR #545, opened 2023-12-01
Instead of allocating the JSON in memory, write to a temp file

adrianclay:
Then upload the temp file to S3

Allocating the full attachment in memory was blowing the heap size when an attachment was 100Mb.

## What

## Why

## Type of change

## Checklist:


adrianclay:
While this does reduce the amount of memory used in one part of the process, there are multiple places where large strings are allocated onto the heap. 😿 This change by itself didn't make an impactful change to the minimum RAM requirements.

Closing for now, but can easily re-open when we come to address Large Attachments again.

PR #570, opened 2024-01-08
Update Gradle to 7.4

adrianclay:
## What

Gradle update to 7.4

Also remove some unused gradle wrapper code from root of repo.

## Why

Current gradle versions don't support Java 17. Java 17 is the next Long Term Support version of Java. Updating Gradle removes a barrier to updating java.

## Type of change

## Checklist:


adrianclay:
Superceeds #564

ole4ryb:
in ..\smoke-tests\gradle\wrapper\gradle-wrapper.properties we use ver. 7.4.2. Maybe we can upgrade everything to ver. 7.4 to it is consistent across the whole application ?

e2e-tests/gradlew
```diff
@@ -17,67 +17,101 @@
...
-SAVED="`pwd`"
-cd "`dirname \"$PRG\"`/" >/dev/null
-APP_HOME="`pwd -P`"
-cd "$SAVED" >/dev/null
+
+APP_HOME=$( cd "${APP_HOME:-./}" && pwd -P ) || exit
 
 APP_NAME="Gradle"
-APP_BASE_NAME=`basename "$0"`
+APP_BASE_NAME=${0##*/}
```
ole4ryb, line 86: why are we replacing `basename "$0"` with ${0##*/} ?

adrianclay: Sounds like a question for the Gradle wrapper maintainers, given this file is their copyright, but you can find [more information on that syntax here](https://unix.stackexchange.com/questions/214465/what-does-prog-0-mean-in-a-bash-script).

ole4ryb (approved):

PR #580, opened 2024-01-22
Remove unused snakeyaml dependency

adrianclay:
## Why

This dependency is out of date, but more importantly it doesn't appear to being used anywhere.

Following on from #578 

## Type of change

## Checklist:

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -34,7 +34,6 @@ dependencies {
 	implementation 'org.springframework.boot:spring-boot-starter-data-mongodb'
 	implementation 'org.springframework:spring-jms:5.3.31'
 
-	implementation 'org.yaml:snakeyaml:1.33'
 	implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.4.2'
 
 	// WebClient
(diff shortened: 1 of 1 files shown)


service/build.gradle
```diff
@@ -34,7 +34,6 @@ dependencies {
 	implementation 'org.springframework.boot:spring-boot-starter-data-mongodb'
 	implementation 'org.springframework:spring-jms:5.3.31'
 
-	implementation 'org.yaml:snakeyaml:1.33'
```
ole4ryb, line 37: I am a bit confused as snakeyaml is used by org.springframework.boot:spring-boot-starter ver. 2.7.18. If we remove ver. 1.33 (with 1 vulnerability) then we got to ver. 1.30 with 7 vulnerabilities. I can see that org.springframework.boot:spring-boot-starter depends on snakeyaml so even if we remove it from here we drop to ver. 1.30.

adrianclay: So `implementation` declares a primary dependency on a package. A primary dependency being on your code relies on explicitly.

We do have a secondary dependency on `snakeyaml` via `spring-boot-starter`.

We don't have any explicit dependency on `snakeyml`, in other words, if `spring-boot-starter` decided to no longer need `snakeyml`, by replacing it with a different YML parser, then our software would continue to compile work.

Line 95 is the one which tells Gradle that when resolving dependencies, make sure to pin the `snakeyml` version to 1.33.

Hope this makes sense!

PR #608, opened 2024-02-21
Remove dependency override for testcontainers

adrianclay:
## Why

The version we're using already has the specified version of commons-compress

## Type of change

## Checklist:


HarveySembhyNHS (approved):
Approved

PR #612, opened 2024-02-21
Remove json-path downgrade

adrianclay:
## Why

Versioning of reactor netty dependendency is already managed as part of the spring-boot-starter-webflux package, which is in turn versioned as part of Spring Framework version.

By having this explicit dependency we were downgrading the version of json-path in use.

This resolves a high CVE in json-smart

## Type of change

## Checklist:


HarveySembhyNHS (approved):
Approved

PR #691, opened 2024-04-10
Update CHANGELOG.md

azam145:
## What

## Why

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## [2.0.2] - 2024-04-10
-### Added Automatic adaptor docker image build
-### Changed Large message size handling
+
+### Fixed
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.0.2] - 2024-04-10
+### Added Automatic adaptor docker image build
+### Changed Large message size handling
```
adrianclay, line 11: 🤣  I'm not sure what format you're using here, but it certainly isn't the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) formatting!

Also, the changelog is aimed at end users. They aren't going to get much information from this information.

There isn't that much that users would care about in this release, but there were some package updates, so I'd recommend putting something like...

```suggestion
### Fixed
- Updated dependencies to keep adaptor secure.
```

adrianclay (approved):

PR #729, opened 2024-05-07
Add scenario for '3 Attachments with 2 Absent' to wiremock

MartinWheelerMT:
## What

Add scenario for '3 Attachments with 2 Absent' to wiremock

## Why

We need to test that all types of Absent Attachment are handled correctly and to be able to compare that they are all dealt with correctly.  We have included attachment types where:

* The document reference points to a file which does not exist
* The document reference is missing a URL
* The document reference is a valid attachment

## Type of change

## Checklist:


MartinWheelerMT:
Incorrect PR - Closing

PR #776, opened 2024-07-01
NIAD-3050: Preserve UUIDs when mapping

MartinWheelerMT:
## What

Updated the RandomIdGeneratorService and IdMapper to preserve UUIDs if a valid UUID is passed in.
Added additional testing for these services
Update unit tests to add mocking for the additional method within RandomIdGeneratorService

## Why

Currently the GP2GP Sending adaptor does not preserve the UUIDs for resources when mapping from Json to XML.  Valid GUID's should be preserved for resources where possible.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+* When mapping resources, if a UUID identifier is provided, this will be preserved in the produced XML. If a non-UUID identifier is provided, a new UUID will continue to be generated.
+
 ## [2.0.4] - 2024-06-17
 
 ### Fixed
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
@@ -21,12 +21,11 @@ import uk.nhs.adaptors.gp2gp.ehr.exception.EhrMapperException;
 /**
  * There exists no requirement within GP Connect FHIR specification that the `IdType`
  * field is populated with a UUID.
- * The GP2GP HL7 specification however does mandate that DCE UUIDs are used within the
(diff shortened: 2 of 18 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
```diff
@@ -39,23 +38,43 @@ public class IdMapper {
...
     }
 
-    public String getOrNew(Reference reference, Boolean isResourceMapped) {
+    public String getOrNewFromReference(Reference reference) {
         if (NOT_ALLOWED.contains(reference.getReferenceElement().getResourceType())) {
             throw new EhrMapperException("Not allowed to use agent-related resource with IdMapper");
         }
 
-        MappedId defaultResourceId = new MappedId(randomIdGeneratorService.createNewId(), isResourceMapped);
+        var referenceId = reference.getReferenceElement().getIdPart();
```
ole4ryb, line 57: I think introducing a variable for reference.getReferenceElement() would give more readability.

ole4ryb:
I can't see any entries in CHANGELOG file.

MartinWheelerMT:
> I can't see any entries in CHANGELOG file.

Yeah I haven't added it yet, I was testing A > B > A First, but it will be added shortly.  For some reason the build wouldn't trigger on Jenkins when it was a draft request!

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/IdMapper.java
MartinWheelerMT: Fair point, will update that now

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+* When mapping resources, if a UUID identifier is provided, this will preserved in the produced XML.  If a non-UUID
```
ole4ryb, line 11: I think there is a 'be' missing after "will".

ole4ryb (approved):

PR #812, opened 2024-08-07
NIAD-3005: Add confidentiality service

MartinWheelerMT:
## What

* Add ConfidentialityService to provide "NOPAT" `confidentialityCode` when provided with a "NOPAT" `meta.security` and the `RedactionsContext` is set to "RCMR_IN030000UK06"
* Add unit tests for ConfidentialityService

## Why

As part of the Redaction Changes it is required to detect if:

* A message is a redaction message (`RCMR_IN030000UK07`)
* The provided `meta.security` is a `NOPAT` Coding, such as below:

```json
{
    "security": [
        {
            "system":"http://hl7.org/fhir/v3/ActCode",
            "code":"NOPAT",
            "display":"no disclosure to patient, family or caregivers without attending provider's authorization"
        }
    ]
}
```

If so, the `ConfidentialityService` will return the following XML string to be added into requesting resources:

```xml
<confidentialityCode
    code="NOPAT"
    codeSystem="2.16.840.1.113883.4.642.3.47"
    displayName="no disclosure to patient, family or caregivers without attending provider's authorization"
/>
```

## Type of change

## Checklist:


service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+    );
+    private static final List<Coding> NON_NOPAT_META_SECURITY = List.of(
+        new Coding(
+            "http://hl7.org/fhir/v3/ActCode",
+            "NON-NOPAT",
+            "random disclosure to patient, family or caregivers without attending provider's authorization"
+        )
+    );
+    public static final String NON_REDACTION_INTERACTION_ID = "RCMR_IN030000UK06";
+    public static final String REDACTION_INTERACTION_ID = "RCMR_IN030000UK07";
```
martin-nhs, line 34: Would it be worth having these within some externalised test utility, I can foresee this being handy within other tests within the future!

Although, arguably, my suggestion may be going against YAGNI 😆

service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+@ExtendWith(MockitoExtension.class)
+public class ConfidentialityServiceTest {
+    private static final List<Coding> NO_META_SECURITY = List.of(new Coding());
+    private static final List<Coding> NOPAT_META_SECURITY = List.of(
+        new Coding(
+            "http://hl7.org/fhir/v3/ActCode",
+            "NOPAT",
+            "no disclosure to patient, family or caregivers without attending provider's authorization"
+        )
+    );
```
martin-nhs, line 25: Rather than `list.of()`, we could use `Collections.singletonList()` here 😄

service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final List<Coding> NON_NOPAT_META_SECURITY = List.of(
+        new Coding(
+            "http://hl7.org/fhir/v3/ActCode",
+            "NON-NOPAT",
+            "random disclosure to patient, family or caregivers without attending provider's authorization"
+        )
+    );
```
martin-nhs, line 32: Same as above (`singletonList()`)!

service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityService.java
```diff
@@ -0,0 +1,46 @@
...
+    private static final String REDACTION_CONFIDENTIALITY_CODE = """
+        <confidentialityCode
+            code="NOPAT"
+            codeSystem="2.16.840.1.113883.4.642.3.47"
+            displayName="no disclosure to patient, family or caregivers without attending provider's authorization"
+        />""";
+
+    private final RedactionsContext redactionsContext;
+
+    public Optional<String> generateConfidentialityCode(List<Coding> metaSecurity) {
```
adrianclay, line 27: I don't have strong opinions on this, but I wonder if passing in the whole resource would make the API a little nicer from the calling side.

i.e. if the method signature was

```suggestion
    public Optional<String> generateConfidentialityCode(Resource resource) {
```

Then when called, the could would be:

```java
    generateConfidentialityCode(observation);
```

vs

```java
    generateConfidentialityCode(observation.getMeta().getSecurity());
```

service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
MartinWheelerMT: Hmm is there a particular reason for `Collections.SingletonList()`, most places I've read seem to thing that they are pretty much the same, but you should use ``List.of() as it is a newer language feature, such as below:

"_If you can use a more recent Java version, it's preferable for its familiarity, clarity and because we use List interface much more than Collections. If restricted to an older Java version, Collections.singletonList remains a solid choice._"

"_Both Collections::singletonList and List:of are great choices for creating single-element lists. If you're fortunate enough to be using a version of Java that supports both methods (9 and above), then I recommend going with List:of for its ease of use, readability, and better-documented immutability._"

service/src/main/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityService.java
MartinWheelerMT: I can do this

service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
MartinWheelerMT: Have moved these into the RedactionsContext

service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
MartinWheelerMT: I'll go with it because it make little odds and I'm making the other changes anyway!

service/src/test/java/uk/nhs/adaptors/gp2gp/common/service/ConfidentialityServiceTest.java
MartinWheelerMT: I'll go with it because it make little odds and I'm making the other changes anyway!

adrianclay (approved):

PR #842, opened 2024-08-23
NIAD-3137: EMIS rejects flat consultation structure from GP2GP Adaptor

MartinWheelerMT:
## What

* Update test files for EncounterComponentsMapperTest to now deal with cases where a virtual category is added for topics without a category. 
* Add functionality to EncounterComponentsMapper to now deal with cases where a virtual category is added for topics without a category. 
* Refactor EncounterComponentsMapper for readability. 

## Why

When attempting to transfer a patient containing a "flat" consultation (where there is a "TOPIC" but with no "CATEGORY") then EMIS reject the patient transfer as they _"expect all TOPIC's to contain a CATEGORY"_.

To avoid the transfer failing, we should create a "virtual" category to contain the resources when translating.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* When mapping consultations which are "flat" (i.e., they contain a `TOPIC` without a `CATEGORY`) we now wrap the 
+  resource into a virtual `CATEGORY`.
 * Added a scheduled delay checker to update EhrExtract to "Integration Failure" state if sentAt exceeds 8 days and no acknowledgment is received. 
 
 ### Added
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
@@ -418,7 +418,7 @@ public class EncounterComponentsMapper {
 
         var params = CompoundStatementParameters.builder()
             .nested(nested)
-            .id(messageContext.getIdMapper().newId(ResourceType.List, topicList.getIdElement()))
+            .id(messageContext.getIdMapper().getOrNew(ResourceType.List, topicList.getIdElement()))
             .classCode(classCode)
             .compoundStatementCode(compoundStatementCode)
             .statusCode(COMPLETE_CODE)
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
@@ -128,28 +128,73 @@ public class EncounterComponentsMapper {
     }
 
     private String mapTopicListComponents(ListResource topicList) {
-        return mapCategorizedResources(topicList) + mapListResourceToComponents(topicList);
+        return String.join(
+            StringUtils.EMPTY,
+            mapCategorizedResources(topicList),
+            mapContainedResources(topicList),
+            mapUncategorizedResources(topicList)
+        );
     }
 
     private String mapCategorizedResources(ListResource topicList) {
         return topicList.getEntry().stream()
             .map(EncounterComponentsMapper::getReferenceElement)
-            .filter(reference -> reference.getValue().matches(LIST_REFERENCE_PATTERN))
+            .filter(reference -> reference
+                .getValue()
+                .matches(LIST_REFERENCE_PATTERN)
+            )
             .map(this::getRequiredResource)
-            .filter(resource -> ResourceType.List.equals(resource.getResourceType()))
             .map(ListResource.class::cast)
-            .filter(listResource -> CodeableConceptMappingUtils.hasCode(listResource.getCode(), List.of(CATEGORY_LIST_CODE)))
             .map(this::mapCategoryListToComponent)
             .collect(Collectors.joining());
     }
 
     private String mapCategoryListToComponent(ListResource categoryList) {
+        if (!CodeableConceptMappingUtils.hasCode(categoryList.getCode(), List.of(CATEGORY_LIST_CODE))) {
+            throw new EhrMapperException(
+                    "Unexpected list %s referenced in Consultation, expected list to be coded as Category (EHR) or be a container"
+                            .formatted(categoryList.getId())
+            );
+        }
+
         return buildCompoundStatement(
             categoryList,
             CATEGORY.getCode(),
             prepareCdForCategory(categoryList),
             true,
             mapListResourceToComponents(categoryList)
+        );
+    }
+
+    private String mapContainedResources(ListResource topicList) {
+        return topicList.getEntry().stream()
+            .filter(entry -> getReferenceElement(entry)
+                .getValue()
+                .matches(CONTAINED_RESOURCE_REFERENCE_PATTERN)
+            )
+            .map(this::mapItemToComponent)
+            .flatMap(Optional::stream)
+            .collect(Collectors.joining());
+    }
+
+    private String mapUncategorizedResources(ListResource topicList) {
+        var components = topicList.getEntry().stream()
+            .map(EncounterComponentsMapper::getReferenceElement)
+            .filter(reference ->
+                !reference.getValue().matches(LIST_REFERENCE_PATTERN)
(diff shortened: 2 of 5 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/NonConsultationResourceMapper.java
```diff
@@ -1,6 +1,7 @@
 package uk.nhs.adaptors.gp2gp.ehr.mapper;
 
 import com.github.mustachejava.Mustache;
+import io.micrometer.common.util.StringUtils;
```
adrianclay, line 4: Strange import.

MartinWheelerMT: Fixed, must have selected the incorrect import in error.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
```diff
@@ -94,240 +95,260 @@ public class EncounterComponentsMapper {
...
     private Optional<String> mapItemToComponent(ListResource.ListEntryComponent item) {
         final String referenceValue = item.getItem().getReference();
         LOGGER.debug("Processing list item {}", referenceValue);
 
         // TODO: workaround for NIAD-1409 where text appears instead of a resource reference
         if (referenceValue == null || NIAD_1409_INVALID_REFERENCE.equals(referenceValue)) {
-            LOGGER.warn("Detected an invalid reference in the GP Connect Demonstrator dataset. "
-                    + "Skipping resource with reference=\"{}\" display=\"{}\"", referenceValue,
-                item.getItem().getDisplay());
+            LOGGER.warn(NIAD_1490_INVALID_REFERENCE_MESSAGE, referenceValue, item.getItem().getDisplay());
```
adrianclay, line 221: Good-ness-me.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* Added a scheduled delay checker to update EhrExtract to "Integration Failure" state if sentAt exceeds 192 hours and 
+no acknowledgment is received.
```
adrianclay, line 12: Dodgy merge
```suggestion
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
```diff
@@ -94,99 +93,90 @@ public class EncounterComponentsMapper {
...
-            .collect(Collectors.toList());
+        return messageContext
+            .getInputBundleHolder()
+            .getListReferencedToEncounter(encounter.getIdElement(), CONSULTATION_LIST_CODE)
+            .map(this::mapListResourceToComponent)
+            .orElse(StringUtils.EMPTY);
+    }
 
-        return topics.stream()
+    private String mapListResourceToComponent(ListResource listResource) {
```
adrianclay, line 103: This looks like `mapConsultationListResourceToComponent` (if I'm understanding this) as a more specific name? List is a bit vague.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
```diff
@@ -94,99 +93,90 @@ public class EncounterComponentsMapper {
...
+            mapTopicListComponents(topicList)
+        );
     }
 
     private String mapTopicListComponents(ListResource topicList) {
+        return String.join(
+            StringUtils.EMPTY,
+            mapCategorizedResources(topicList),
+            mapContainedResources(topicList),
+            mapUncategorizedResources(topicList)
```
adrianclay, line 133: So from my understanding of this code, it's been introduced to handle resources which have been referenced in each of the two different forms:
```
List/abc#def
Observation/ghi
```

AFAIK is a low level implementation detail of how FHIR resources can live within the Bundle, and ideally if there is a method which takes in either of those Reference types, it should be able to return the `Resource` which it referrs to.

Furthermore, for this ticket it looks like we've fixed the situation where the resource entries are all of the form `Observation/xxx`, however if we had a mix and match of both forms of reference then the resources of the "contained" form won't be wrapped in the the Topic CompoundStatement.

I believe that our business logic should be agnostic of the type of Reference being used within the list, and instead should care more about the _type_ of Resource that the Reference is pointing at.

This method to me should have two cases:
1. I've been given a list of all topics, do X
2. I've been given a list of all non-list resources (irrespective of how they're referenced), do Y

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
```diff
@@ -94,99 +93,90 @@ public class EncounterComponentsMapper {
...
-            .map(reference -> messageContext
-                .getInputBundleHolder()
-                .getRequiredResource(reference))
-            .filter(resource -> ResourceType.List.equals(resource.getResourceType()))
+    private String mapCategorizedResources(ListResource topicList) {
+        return topicList.getEntry().stream()
+            .map(EncounterComponentsMapper::getReferenceElement)
+            .filter(reference -> reference
+                .getValue()
+                .matches(LIST_REFERENCE_PATTERN)
```
adrianclay, line 142: Also, our approach to identifying whether something is a LIST or not is janky, because it assumes that the resource isn't of the `contained` form. HOWEVER there is nothing stopping a GP System from providing `List (Heading)` as a contained reference. This defect is separate to this original ticket, but maybe you could a Jira defect for it. 🤣

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
```diff
@@ -380,7 +418,7 @@ private String buildCompoundStatement(
 
         var params = CompoundStatementParameters.builder()
             .nested(nested)
-            .id(messageContext.getIdMapper().getOrNew(ResourceType.List, topicList.getIdElement()))
```
adrianclay, line 383: Hmm, what was the thinking with this change?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
```diff
@@ -128,22 +128,35 @@ private String mapTopicListToComponent(ListResource topicList) {
...
             .map(ListResource.class::cast)
-            .filter(listResource -> CodeableConceptMappingUtils.hasCode(listResource.getCode(), List.of(CATEGORY_LIST_CODE)))
             .map(this::mapCategoryListToComponent)
             .collect(Collectors.joining());
     }
 
     private String mapCategoryListToComponent(ListResource categoryList) {
+        if (!CodeableConceptMappingUtils.hasCode(categoryList.getCode(), List.of(CATEGORY_LIST_CODE))) {
+            throw new EhrMapperException(
+                    "Unexpected list %s referenced in Consultation, expected list to be coded as Category (EHR) or be a container"
```
adrianclay, line 155: I think the exception could be improved. The container part seems strange to me as we already know it's a list. Also being clear that we found this resource as an entry within a Topic list, might help clarify what the situation was.

From:
"Unexpected list NaughtyList referenced in Consultation, expected list to be coded as Category (EHR) or be a container"

To:
"List resource with id "NaughtyList" as entry within List (Topic) is required to be coded as List (Category) 24781000000107 to match GP Connect Specification, but no such code found on resource."

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapper.java
adrianclay: I've had a further look into this and the behaviour for "contained" resources includes the following requirement:
> The contained resource can’t be referred to outside the main resource, not even from another resource in the same bundle.

So therefore a resource in the form `List/abc#def` is invalid. The only valid way of referencing a contained resource is with the `#blah` form.

adrianclay (changes requested):
UNACCEPTABLE. Please reconsider your choices.

adrianclay (approved):

PR #849, opened 2024-09-01
Migrate from v1 to v2 of AWS SDK for Java

adrianclay:
## Why

v1 [stops getting updates](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/) as of Dec 2025

## Type of change

Internal change (non-breaking change with no effect on the functionality affecting end users)

## Checklist:

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -133,7 +133,6 @@ tasks.withType(Test) {
 }
 
 check.dependsOn integrationTest
-jacocoTestReport.dependsOn integrationTest
 
 spotbugsTest.enabled = false
 spotbugsIntegrationTest.enabled = false


service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/StorageConnectorConfiguration.java
```diff
@@ -16,6 +16,8 @@
 @Configuration
 @ConfigurationProperties(prefix = "gp2gp.storage")
 public class StorageConnectorConfiguration {
+
+    private static final Region REGION = Region.EU_WEST_2;
```
adrianclay, line 20: Dodgy.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/configuration/CustomTrustStoreTest.java
```diff
@@ -0,0 +1,69 @@
...
+    static void tearDown() {
+        s3Mock.shutdown();
+    }
+
+    @Test
+    void trustManagerLoadsSuccessfullyTest() throws NoSuchFieldException, IllegalAccessException {
+
+        Field s3ClientField = CustomTrustStore.class.getDeclaredField("s3Client");
+        s3ClientField.setAccessible(true);
+        s3ClientField.set(customTrustStore, s3Client);
```
adrianclay, line 61: Personally not a fan of messing around with reflection to create tests. Any way to make the `s3Client` a constructor argument?

service/src/main/java/uk/nhs/adaptors/gp2gp/common/storage/S3StorageConnector.java
```diff
@@ -2,44 +2,41 @@
...
+    private final S3Client s3client;
     private final String bucketName;
 
-    protected S3StorageConnector(StorageConnectorConfiguration configuration) {
+    protected S3StorageConnector(S3Client s3client, StorageConnectorConfiguration configuration) {
         this.bucketName = configuration.getContainerName();
-        this.s3client = AmazonS3ClientBuilder
-            .standard()
-            .build();
+        this.s3client = s3client;
```
adrianclay, line 18: Will this use the Bean defined as `getS3Client` inside of `StorageConnectorConfiguration.java` ? Because if so, then that bean only gets constructed when a truststore is defined. Whereas previously this was being constructed all the time.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/configuration/CustomTrustStoreTest.java
adrianclay: Thinking about it, I'd probably make the `AppInitializer` construct it's own `CustomTrustStore` instead of relying on dependency injection. Also moving into `AppInitializer` the logic to optionally construct the `S3Client` that is currently inside of `StorageConnectorConfiguration`.

`AppInitializer` can instead depend on `StorageConnectorConfiguration`.

.github/workflows/build_workflow.yml
```diff
@@ -109,9 +109,9 @@ jobs:
       - name: Execute Unit Tests
         run: |
           if [[ $GITHUB_ACTOR == 'dependabot[bot]' ]]; then
-            ./gradlew test jacocoTestReport --parallel --build-cache
+            ./gradlew test jacocoTestReport -x integrationTest --parallel --build-cache
           else
-            ./gradlew test jacocoTestReport sonar --parallel --build-cache
+            ./gradlew test jacocoTestReport sonar -x integrationTest --parallel --build-cache
```
adrianclay, line 114: I'd move this stuff into a separate PR.

service/build.gradle
```diff
@@ -132,6 +133,7 @@ tasks.withType(Test) {
 }
 
 check.dependsOn integrationTest
+jacocoTestReport.dependsOn integrationTest
```
adrianclay, line 136: Is this not what's causing the integration tests to run as part of the unit tests?

PR #864, opened 2024-09-11
NIAD-2840: Identifies no acknowledgment after 8 Days

ole4ryb:
## What
Scheduler delay checker was added.

## Why

The scheduler delay checker now marks transfer requests as failed if they exceed the 8-day limit without receiving a final acknowledgment.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/Gp2gpConfiguration.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/Gp2gpConfiguration.java
@@ -5,10 +5,12 @@ import org.springframework.context.annotation.Configuration;
 
 import lombok.Getter;
 import lombok.Setter;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 @Getter
 @Setter
 @Configuration
+@EnableScheduling
 @ConfigurationProperties(prefix = "gp2gp")
 public class Gp2gpConfiguration {
     private int largeAttachmentThreshold;
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
@@ -10,15 +10,18 @@ import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.mongodb.core.FindAndModifyOptions;
 import org.springframework.data.mongodb.core.MongoTemplate;
 import org.springframework.data.mongodb.core.query.Criteria;
 import org.springframework.data.mongodb.core.query.Query;
 import org.springframework.data.mongodb.core.query.Update;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
 
 import com.mongodb.client.result.UpdateResult;
@@ -115,12 +118,102 @@ public class EhrExtractStatusService {
     private static final String FILENAME_TYPE_PLACEHOLDER = "FILENAME_PLACEHOLDER_ID=";
     private static final String ACKS_SET = ACK_HISTORY + DOT + ACKS;
     public static final int EHR_EXTRACT_SENT_DAYS_LIMIT = 8;
-    private static final String REASON_ERROR_CODE = "04";
+    private static final String REASON_ERROR_CODE = "99";
     public static final String REASON_ERROR_MESSAGE = "The acknowledgement has been received after 8 days";
 
     private final MongoTemplate mongoTemplate;
     private final EhrExtractStatusRepository ehrExtractStatusRepository;
     private final TimestampService timestampService;
+
+    @Scheduled(cron = "${timeout.cronTime}")
+    public void checkForEhrExtractAckTimeouts() {
+        List<EhrExtractStatus> inProgressEhrExtractTransfers = findInProgressTransfers();
+        var ehrExtractStatusWithExceededUpdateLimit = inProgressEhrExtractTransfers
+            .stream()
+            .filter(ehrExtractStatus -> Objects.isNull(ehrExtractStatus.getEhrReceivedAcknowledgement()))
+            .filter(this::hasLastUpdateExceededEightDays)
+            .toList();
+
+        if (!ehrExtractStatusWithExceededUpdateLimit.isEmpty()) {
+            updateEhrExtractStatusListWithEhrReceivedAcknowledgementError(ehrExtractStatusWithExceededUpdateLimit,
+                                                                          REASON_ERROR_CODE,
+                                                                          REASON_ERROR_MESSAGE);
+        }
+    }
+
+    boolean hasEhrStatusReceivedAckWithErrors(String conversationId) {
+
+        var ehrExtractStatus = fetchEhrExtractStatus(conversationId, "NACK");
+
+        var ehrReceivedAcknowledgement = ehrExtractStatus.getEhrReceivedAcknowledgement();
+        if (ehrReceivedAcknowledgement == null) {
+            return false;
+        }
+
+        var errors = ehrReceivedAcknowledgement.getErrors();
+        if (errors == null || errors.isEmpty()) {
+            return false;
+        }
+
+        var ehrReceivedAckErrorDetailsThatContainsSearchedErrCodeAndMsg = errors.stream()
+            .filter(err -> REASON_ERROR_CODE.equals(err.getCode()) && REASON_ERROR_MESSAGE.equals(err.getDisplay()))
+            .findAny();
+
+        return ehrReceivedAckErrorDetailsThatContainsSearchedErrCodeAndMsg.isPresent();
+    }
+
+    public void updateEhrExtractStatusListWithEhrReceivedAcknowledgementError(List<EhrExtractStatus> ehrExtractStatusList,
+                                                                              String errorCode,
+                                                                              String errorMessage) {
+
+        ehrExtractStatusList.stream().forEach(ehrExtractStatus -> {
+            try {
+                updateEhrExtractStatusWithEhrReceivedAckError(ehrExtractStatus.getConversationId(), errorCode, errorMessage);
+            } catch (Exception e) {
+                logger().error("An error occurred when closing a failed process for conversation_id: {}",
+                               ehrExtractStatus.getConversationId(), e);
+            }
+        });
+    }
+
+    protected EhrExtractStatus updateEhrExtractStatusWithEhrReceivedAckError(String conversationId,
+                                                                             String errorCode,
+                                                                             String errorMessage) {
+
+        Update update = createUpdateWithUpdatedAt();
+        update.addToSet(RECEIVED_ACK_ERRORS, ErrorDetails.builder().code(errorCode).display(errorMessage).build());
+
+        EhrExtractStatus ehrExtractStatus = mongoTemplate.findAndModify(
+            createQueryForConversationId(conversationId),
+            update,
+            getReturningUpdatedRecordOption(),
+            EhrExtractStatus.class);
+
+        if (ehrExtractStatus == null) {
+            throw new EhrExtractException(format(
+                "Couldn't update EHR received acknowledgement with error information because EHR status doesn't exist, conversation_id: %s",
+                conversationId));
+        }
+
+        logger().info("EHR status (EHR received acknowledgement) record successfully "
+                      + "updated in the database with error information conversation_id: {}", conversationId);
+
+        return ehrExtractStatus;
+    }
+
+    public boolean hasLastUpdateExceededEightDays(EhrExtractStatus ehrExtractStatus) {
+
+        var now = Instant.now();
+
+        if (ehrExtractStatus.getEhrExtractCorePending() == null) {
+            return false;
+        }
+
+        Duration duration = Duration.between(ehrExtractStatus.getEhrExtractCorePending().getSentAt(), now);
+        long daysSinceLastUpdate = duration.toDays();
+
+        return daysSinceLastUpdate > EHR_EXTRACT_SENT_DAYS_LIMIT;
+    }
 
     public void saveEhrExtractMessageId(String conversationId, String messageId) {
         Optional<EhrExtractStatus> ehrExtractStatusOptional = ehrExtractStatusRepository.findByConversationId(conversationId);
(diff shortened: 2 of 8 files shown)


martin-nhs:
Looks like the `CHANGELOG.md` needs updated? This may have already been done with the last merge, can't remember!

service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/Gp2gpConfiguration.java
```diff
@@ -5,10 +5,12 @@
 
 import lombok.Getter;
 import lombok.Setter;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 @Getter
 @Setter
 @Configuration
+@EnableScheduling
```
martin-nhs, line 13: I would separate this out into it's own configuration - although not required, it helps adhere to the single responsibility principle.

Example of how I usually set it up in other projects:

```java
@Configuration
@EnableScheduling
class ScheduleConfiguration {
 // Empty for now, though would allow for scheduling specific bean registrations.
}
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +118,103 @@ public class EhrExtractStatusService {
     private static final String FILENAME_TYPE_PLACEHOLDER = "FILENAME_PLACEHOLDER_ID=";
     private static final String ACKS_SET = ACK_HISTORY + DOT + ACKS;
     public static final int EHR_EXTRACT_SENT_DAYS_LIMIT = 8;
-    private static final String REASON_ERROR_CODE = "04";
+    private static final String REASON_ERROR_CODE = "99";
```
martin-nhs, line 121: Out of scope for this PR possibly, though wouldn't it be better to separate this out into an enumeration of sorts? I could see the benefit of having a central location for these error codes I see in this codebase.

Furthermore, a little comment to describe what these mean could be useful! Not a clue what `99` means to us in this context! 😄

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +118,103 @@ public class EhrExtractStatusService {
...
+
+        ehrExtractStatusList.stream().forEach(ehrExtractStatus -> {
+            try {
+                updateEhrExtractStatusWithEhrReceivedAckError(ehrExtractStatus.getConversationId(), errorCode, errorMessage);
+            } catch (Exception e) {
+
+                logger().error("An error occurred when closing a failed process for conversation_id: {}",
+                               ehrExtractStatus.getConversationId(), e);
+                throw e;
+            }
```
martin-nhs, line 178: Are there any more specific exceptions we can catch here rather than using `Exception.java`? Furthermore could we please rename `e` to `exception` 😄

Small, but it all adds up imo!

service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/Gp2gpConfiguration.java
ole4ryb: as there were no other beans and configs defined in this place hence I thought that it wouldn't add a big burden if I add this annotation here. As this annotation is declarative and there are no many other such annotations for GP2GPConfiguration.java class I think I'd leave it here for the time being but a good point to watch it in the future if there appear more scheduler-related things.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
ole4ryb: I think it is a good point for the future refactoring effort and I see the benefit in aggregating all the error codes as Enums in one place 👍. Will take a note of it. The error code will be renamed to be more descriptive. 👍

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +118,107 @@ public class EhrExtractStatusService {
     private static final String FILENAME_TYPE_PLACEHOLDER = "FILENAME_PLACEHOLDER_ID=";
     private static final String ACKS_SET = ACK_HISTORY + DOT + ACKS;
     public static final int EHR_EXTRACT_SENT_DAYS_LIMIT = 8;
-    private static final String REASON_ERROR_CODE = "04";
-    public static final String REASON_ERROR_MESSAGE = "The acknowledgement has been received after 8 days";
+    private static final String UNEXPECTED_CONDITION_ERROR_CODE = "99";
```
martin-nhs, line 121: As an aside, is there any documentation regarding what these error codes should be? I'd enjoy reading up on them as I don't think I've spotted anything about these so far! 😄

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +118,107 @@ public class EhrExtractStatusService {
...
+    }
+
+    boolean hasEhrStatusReceivedAckWithErrors(String conversationId) {
+
+        var ehrExtractStatus = fetchEhrExtractStatus(conversationId, "NACK");
+
+        var ehrReceivedAcknowledgement = ehrExtractStatus.getEhrReceivedAcknowledgement();
+        if (ehrReceivedAcknowledgement == null) {
+            return false;
+        }
```
martin-nhs, line 152: I'm seeing quite a lot of null values being handled in this class, is there any way we can use `Optional<T>` here to reduce the risk of NPEs?

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +118,107 @@ public class EhrExtractStatusService {
...
+            .toList();
+
+        if (!ehrExtractStatusWithExceededUpdateLimit.isEmpty()) {
+            updateEhrExtractStatusListWithEhrReceivedAcknowledgementError(ehrExtractStatusWithExceededUpdateLimit,
+                                                                          UNEXPECTED_CONDITION_ERROR_CODE,
+                                                                          UNEXPECTED_CONDITION_ERROR_MESSAGE);
+        }
+    }
+
+    boolean hasEhrStatusReceivedAckWithErrors(String conversationId) {
```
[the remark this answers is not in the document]
ole4ryb: for me it sounds quite descriptive but I've adjusted it to be even more descriptive :)

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
ole4ryb: I couldn't find anything in our codebase either. I think it might be somewhere described in NHS documentation.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -622,4 +711,8 @@ private EhrExtractStatus fetchEhrExtractStatus(String conversationId, String mes
         return ehrExtractStatusRepository.findByConversationId(conversationId)
             .orElseThrow(() -> new UnrecognisedInteractionIdException(messageType, conversationId));
     }
+
+    protected Logger logger() {
+        return LOGGER;
+    }
```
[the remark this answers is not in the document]
ole4ryb: in this case I thought to encapsulate it as it helped with testing.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +118,107 @@ public class EhrExtractStatusService {
...
 
+    @Scheduled(cron = "${timeout.cronTime}")
+    public void checkForEhrExtractAckTimeouts() {
+        List<EhrExtractStatus> inProgressEhrExtractTransfers = findInProgressTransfers();
+        var now = Instant.now();
+        var ehrExtractStatusWithExceededUpdateLimit = inProgressEhrExtractTransfers
+            .stream()
+            .filter(ehrExtractStatus -> Objects.isNull(ehrExtractStatus.getEhrReceivedAcknowledgement()))
+            .filter(ehrExtractStatus -> hasLastUpdateExceededEightDays(ehrExtractStatus, now))
+            .toList();
```
adrianclay, line 136: What was your thinking for converting it to a list here and then back to a stream inside of `updateEhrExtractStatusListWithEhrReceivedAcknowledgementError`, seems unnecessary.

ole4ryb: It's better to keep `ehrExtractStatusWithExceededUpdateLimit` as a List rather than a Stream in this case. This is because you need to check whether the list is empty, and if you use a Stream, applying findAny() (which is a terminal operation) like below:
`if (!ehrExtractStatusWithExceededUpdateLimit.findAny().isEmpty()) {` ...

will consume the stream and make it unusable for further operations. This can lead to a `java.lang.IllegalStateException` if you try to use the stream afterward. By using a List, you can easily check if it's empty and avoid these issues, ensuring that the method operates without exceptions or side effects.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
ole4ryb: good suggestion Martin, I will take a note of it and do the refactoring in the next PR that I am planning to do after this PR.

martin-nhs:
For a separate PR:
`public static final int EHR_EXTRACT_SENT_DAYS_LIMIT = 8;`

I think this is best to be extracted out into `application.yml`.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
adrianclay: > you need to check whether the list is empty

Citation needed.

service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/Gp2gpConfiguration.java
martin-nhs: That's fine, I still think maybe it's better suited in a separate class - **S**OLID!  😄

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +119,104 @@ public class EhrExtractStatusService {
     private static final String FILENAME_TYPE_PLACEHOLDER = "FILENAME_PLACEHOLDER_ID=";
     private static final String ACKS_SET = ACK_HISTORY + DOT + ACKS;
     public static final int EHR_EXTRACT_SENT_DAYS_LIMIT = 8;
-    private static final String REASON_ERROR_CODE = "04";
-    public static final String REASON_ERROR_MESSAGE = "The acknowledgement has been received after 8 days";
+    private static final String UNEXPECTED_CONDITION_ERROR_CODE = "99";
```
martin-nhs, line 122: For a separate PR - it could be good to see if we can centralise these codes somehow, even via an enumeration.

ole4ryb: yes, I suggestions, I've taken a note of it and address it in a separate follow-up PR

martin-nhs (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -115,13 +119,103 @@ public class EhrExtractStatusService {
...
+            .stream()
+            .filter(ehrExtractStatus -> Objects.isNull(ehrExtractStatus.getEhrReceivedAcknowledgement()))
+            .filter(ehrExtractStatus -> hasLastUpdateExceededEightDays(ehrExtractStatus, now));
+
+        updateEhrExtractStatusListWithEhrReceivedAcknowledgementError(ehrExtractStatusWithExceededUpdateLimit,
+                                                                      UNEXPECTED_CONDITION_ERROR_CODE,
+                                                                      UNEXPECTED_CONDITION_ERROR_MESSAGE);
+    }
+
+    boolean hasEhrStatusReceivedAckWithUnexpectedConditionErrors(String conversationId) {
```
adrianclay, line 143:
```suggestion
    boolean hasEhrExtractStatusTimedoutFromNotReceivingAnAcknowledgement(String conversationId) {
```

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -300,20 +394,19 @@ public Optional<EhrExtractStatus> updateEhrExtractStatusContinue(String conversa
...
         if (ack.getErrors() == null && !isEhrStatusWaitingForFinalAck(conversationId)) {
-            LOGGER.warn("Received unexpected acknowledgement of EHR Extract with conversation id=" + conversationId);
+            logger().warn("Received unexpected acknowledgement of EHR Extract with conversation id=" + conversationId);
             return;
         }
 
-        if (hasFinalAckBeenReceived(conversationId)) {
-            LOGGER.warn("Received an ACK message with a conversation_id=" + conversationId + " that is a duplicate");
+        if (hasAcknowledgementExceededEightDays(conversationId, ack.getReceived())
+            && hasEhrStatusReceivedAckWithUnexpectedConditionErrors(conversationId)) {
```
adrianclay, line 402: Once you've renamed `hasEhrStatusReceivedAckWithUnexpectedConditionErrors` to something clearer, then I don't think you need the first check.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -300,20 +394,19 @@ public Optional<EhrExtractStatus> updateEhrExtractStatusContinue(String conversa
...
+
+            logger().warn("Received an ACK message with a conversation_id: {}, but it will be ignored", conversationId);
```
adrianclay, line 404: I'd also be a bit clearer in this warning, to be clearer that the reason its being ignored is because we're already determined that the transfer has timed out because we didn't get an acknowledgement from the requesting practice in time.

PR #865, opened 2024-09-12
NIAD-3173 - Add GitHub Action to replace current GP2GP Jenkins build

martin-nhs:
Due to ongoing challenges with Jenkins, we have initiated the migration to GitHub Actions as part of our strategy to enhance our CI/CD pipeline. This pull request introduces the initial GitHub Actions workflow, `build_workflow.yml`. This workflow is designed to replicate the stages defined in the existing `Jenkinsfile`, ensuring continuity and consistency in our build process.

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -8,9 +8,6 @@ on:
     branches:
       - main
 jobs:
-
-# ------------------------------------------------------------
-
   checkstyle:
     name: Checkstyle
     runs-on: ubuntu-latest
@@ -22,7 +19,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           java-version: 21
-          distribution: 'corretto'
+          distribution: 'temurin'
 
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
@@ -52,7 +49,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           java-version: 21
-          distribution: 'corretto'
+          distribution: 'temurin'
 
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
@@ -71,8 +68,6 @@ jobs:
           path: "./service/build/reports/**"
           compression-level: 9
 
-# ------------------------------------------------------------
-
   unit-tests:
     name: Unit Tests
     runs-on: ubuntu-latest
@@ -85,7 +80,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           java-version: 21
-          distribution: 'corretto'
+          distribution: 'temurin'
 
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
@@ -116,7 +111,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           java-version: 21
-          distribution: 'corretto'
+          distribution: 'temurin'
 
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
@@ -210,7 +205,7 @@ jobs:
       - name: Get End-to-End Test Reports
         run: |
           mkdir e2e-test-reports
-          docker cp e2e-tests:/home/gradle/e2e-tests/build/reports/** ./e2e-test-reports
+          docker cp e2e-tests:/home/gradle/e2e-tests/build/reports ./e2e-test-reports
 
       - name: Dump Docker Logs
         if: always()
@@ -236,9 +231,7 @@ jobs:
           docker network rm commonforgp2gp
         working-directory: ./docker
 
-# ------------------------------------------------------------
-
-  docker:
+  build-and-publish-docker-images:
     name: Build & Publish Docker Images
     runs-on: ubuntu-latest
     needs: [unit-tests, integration-tests, end-to-end-tests]
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - main
+
 jobs:
   checkstyle:
     name: Checkstyle
@@ -323,7 +324,8 @@ jobs:
       - name: Build Docker Image
         run: |
           # Create Docker Tag
-          DOCKER_TAG="${{ secrets.DOCKER_REGISTRY }}/${{ matrix.config.repository }}:$BUILD_ID"
+          DOCKER_REGISTRY="https://${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.${{ secrets.AWS_REGION }}.amazonaws.com"
+          DOCKER_TAG="$DOCKER_REGISTRY/${{ matrix.config.repository }}:$BUILD_ID"
           echo "DOCKER_TAG=$DOCKER_TAG" >> $GITHUB_ENV
 
           # Build Image
@@ -331,7 +333,8 @@ jobs:
 
       - name: Login to AWS ECR
         run: |
-          aws ecr get-login-password --region ${{ secrets.AWS_REGION }} | docker login --username AWS --password-stdin https://${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.${{ secrets.AWS_REGION }}.amazonaws.com
+          DOCKER_REGISTRY="https://${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.${{ secrets.AWS_REGION }}.amazonaws.com"
+          aws ecr get-login-password --region ${{ secrets.AWS_REGION }} | docker login --username AWS --password-stdin $DOCKER_REGISTRY
 
       - name: Publish image to ECR
         run: docker push $DOCKER_TAG
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -158,7 +158,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always()
         with:
-          name: 'Integration Test Reports - Docker Logs'
+          name: 'Integration Test Reports & Docker Logs'
           path: |
             ./service/build/reports/**
             ./scripts/logs/**
@@ -207,8 +207,9 @@ jobs:
           docker compose -f docker-compose.yml -f docker-compose-e2e-tests.yml up --exit-code-from mongodb activemq gp2gp wiremock gpcc gp2gp-e2e-tests
         working-directory: ./docker
 
-      - name: Copy End-to-End Tests Report
-        run: docker cp e2e-tests:/home/gradle/e2e-tests/build .
+      - name: Get End-to-End Test Reports
+        run: docker cp e2e-tests:/home/gradle/e2e-tests/build/reports/** .
+        working-directory: ./service/build/reports
 
       - name: Dump Docker Logs
         if: always()
@@ -221,7 +222,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always()
         with:
-          name: 'End-to-End Test Results - Docker Logs'
+          name: 'End-to-End Test Results & Docker Logs'
           path: |
             ./service/build/reports/**
             ./scripts/logs/**
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -103,7 +103,7 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
 
       - name: Execute Unit Tests
-        run: ./gradlew test --daemon --parallel --build-cache
+        run: ./gradlew test --parallel --build-cache
         working-directory: ./service
 
       - name: Collect Artifacts
@@ -162,7 +162,7 @@ jobs:
         working-directory: ./docker
 
       - name: Execute Integration Tests
-        run: ./gradlew integrationTest --daemon
+        run: ./gradlew integrationTest
         working-directory: ./service
 
       - name: Dump Docker Logs
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -25,9 +25,6 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
 
-      - name: Create Temporary Artifacts Directory
-        run: mkdir artifacts
-
       - name: Checkstyle
         run: |
           ./gradlew checkStyleMain --parallel
@@ -35,7 +32,9 @@ jobs:
         working-directory: ./service
 
       - name: Collect Artifacts
-        run: cp -r ./service/build/reports ./artifacts
+        run: |
+          mkdir -p artifacts
+          cp -r ./service/build/reports ./artifacts
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v4
@@ -64,9 +63,6 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
 
-      - name: Create Temporary Artifacts Directory
-        run: mkdir artifacts
-
       - name: Spotbugs
         run: |
           ./gradlew spotbugsMain --parallel
@@ -74,7 +70,9 @@ jobs:
         working-directory: ./service
 
       - name: Collect Artifacts
-        run: cp -r ./service/build/reports ./artifacts
+        run: |
+          mkdir -p artifacts
+          cp -r ./service/build/reports ./artifacts
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,305 @@
...
+          GP2GP_MONGO_DATABASE_NAME: "gp2gp"
+          GP2GP_MHS_OUTBOUND_URL: "http://mock-mhs-adaptor:8081/mock-mhs-endpoint/accept-only"
+          GP2GP_GPC_GET_URL: "http://gpcc:8090/@ODS_CODE@/STU3/1/gpconnect"
+          GP2GP_LARGE_ATTACHMENT_THRESHOLD: "4500000"
+          GP2GP_LARGE_EHR_EXTRACT_THRESHOLD: "4500000"
+          GPC_CONSUMER_SERVER_PORT: "8090"
+          GPC_CONSUMER_SDS_URL: "http://wiremock:8080/spine-directory/"
+          GPC_CONSUMER_SDS_APIKEY: "anykey"
+          GP2GP_LOGGING_LEVEL: DEBUG
+          GPC_CONSUMER_LOGGING_LEVEL: DEBUG
```
martin-nhs, line 139: Investigate if there is an alternative way to get these environment variables imported via a file on the LFS - without using `source`.

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,305 @@
...
+
+      - name: Execute Integration Tests
+        run: ./gradlew integrationTest
+        working-directory: ./service
+
+      - name: Dump Docker Logs
+        if: always()
+        run: |
+          chmod +x dump_docker_logs.sh
+          ./dump_docker_logs.sh
```
martin-nhs, line 154: I feel like using `chmod` here is a little strange, are there any other ways we can avoid this?

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,306 @@
...
+            build-context: .
+          - directory: gpcc-mock
+            repository: gp2gp-gpcc-mock
+            build-context: ./docker/gpcc-mock
+          - directory: gpc-api-mock
+            repository: gp2gp-gpc-api-mock
+            build-context: ./docker/gpc-api-mock
+          - directory: sds-api-mock
+            repository: gp2gp-sds-api-mock
+            build-context: ./docker/sds-api-mock
```
martin-nhs, line 267: This seems confusing, how can we make this better?

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,306 @@
...
+      - name: Setup Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Generate Build ID
+        run: |
+          BUILD_ID=$(python3 ./scripts/tag.py ${{ github.head_ref || github.ref_name }} ${{ github.run_number }} ${{ github.sha }})
+          echo "Generated the build tag: $BUILD_ID"
+          echo "BUILD_ID=$BUILD_ID" >> $GITHUB_ENV
```
martin-nhs, line 289: Setting up Python to run a script feels overkill, let's port `tag.py` to bash?

.github/workflows/build_workflow.yml
adrianclay: Hmm, I'm a little suspicious of the suggestion of using LFS. Not sure how it would apply here.

.github/workflows/build_workflow.yml
adrianclay: You should be able change the permissions of that file inside of version control.

.github/workflows/build_workflow.yml
adrianclay: I wouldn't worry too much about this for now, as it's what's there already for Jenkins. Not like this PR is making the situation any worse.

martin-nhs: Agreed, we can revisit at a later stage for enhancements like this :)

martin-nhs:
WARNING! Your password will be stored unencrypted in /home/runner/.docker/config.json.
Configure a credential helper to remove this warning.

See [https://docs.docker.com/engine/reference/commandline/login/#credentials-store](https://docs.docker.com/engine/reference/commandline/login/#credentials-store).

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+        uses: actions/checkout@v4
+
+      - name: Setup Java 21 LTS
+        uses: actions/setup-java@v4
+        with:
+          java-version: 21
+          distribution: 'temurin'
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
```
ole4ryb, line 26: maybe not in this PR but as a next iteration it would be good to consider including Gradle caching mechanisms which should improve the performance even better. Here is some analysis about the positive caching impact: https://docs.gradle.org/current/userguide/build_cache_performance.html#analyzing_performance_in_build_scans

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+      - name: Collect Artifacts
+        run: cp -r ./service/build/reports ./artifacts
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: 'Checkstyle Reports'
+          path: ./artifacts/**
+          compression-level: 9
```
ole4ryb, line 46: As per this source: [https://github.com/actions/upload-artifact](url) compression-level 9 gives the best compression but not the best speed. 
Could we maybe choose something like 5 or 6 should should give a better balance between speed and compression size? 
image

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+
+      - name: Delete Temporary Artifacts
+        run: rm -r ./artifacts
```
ole4ryb, line 49: Could we use "rm -rf" as in automated environments it is a more preferable options.
If for any reason, "./artifacts" don't exist, the command "rm -r" will fail, "rm -rf" is more tolerant and will carry on.

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Create Temporary Artifacts Directory
+        run: mkdir artifacts
+
+      - name: Spotbugs
+        run: |
+          ./gradlew spotbugsMain
```
ole4ryb, line 72: Similar to the above, I think spotbugsMain and spotbugsTest can run in parallel
`./gradlew spotbugsMain spotbugsTest --parallel`, which should be quicker, at least in theory...

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+          ./gradlew spotbugsTest
+        working-directory: ./service
+
+      - name: Collect Artifacts
+        run: cp -r ./service/build/reports ./artifacts
```
ole4ryb, line 77: Before copying I would suggest adding a small defensive check `mkdir -p ./artifacts`,
it will create a directory in case (for any reason) "./artifacts" doesn't exist.
Also, if for any reason the reports don't get copied, could we again be quite explicit and say it:
`cp -r ./service/build/reports ./artifacts || { echo "Copy failed"; exit 1; }`
Again, this is just a defensive approach and I think would be helpful in case of any issues with the pipeline.

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+          java-version: 21
+          distribution: 'temurin'
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Create Temporary Artifacts Directory
+        run: mkdir artifacts
+
+      - name: Execute Unit Tests
```
ole4ryb, line 110: In my local env `./gradlew clean test` vs. `./gradlew clean test --parallel --build-cache` results in
3m38s against 39s as a lot is being pulled from cache hence I'd suggest using it or at least exploring this change as an option in the next iteration of changes.
image

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: 'Unit Test Reports'
+          path: ./artifacts/**
+          compression-level: 9
+
+      - name: Delete Temporary Artifacts
```
ole4ryb, line 125: I think "Temporary Artifacts Cleanup" sounds slightly better.
Also, would suggest running "rm -r ./artifacts" with "f" as it is more error-tolerant.
`run: rm -rf ./artifacts`

.github/workflows/build_workflow.yml
martin-nhs: Hey @ole4ryb - thanks for the comment - a useful point!
 
All artifact uploads observed in our successful runs of the workflow indicate an average of `0.6s` to run on the best compression level - given this, and due to the nature of the artifacts produced, I think I am happy to proceed with keeping it at the best compression level of `9` for now. What do you think?

.github/workflows/build_workflow.yml
ole4ryb: I think the variable can get imported using ".env" file.

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,340 @@
...
+          GP2GP_GPC_GET_URL: "http://gpcc:8090/@ODS_CODE@/STU3/1/gpconnect"
+          GP2GP_LARGE_ATTACHMENT_THRESHOLD: "4500000"
+          GP2GP_LARGE_EHR_EXTRACT_THRESHOLD: "4500000"
+          GPC_CONSUMER_SERVER_PORT: "8090"
+          GPC_CONSUMER_SDS_URL: "http://wiremock:8080/spine-directory/"
+          GPC_CONSUMER_SDS_APIKEY: "anykey"
+          GP2GP_LOGGING_LEVEL: DEBUG
+          GPC_CONSUMER_LOGGING_LEVEL: DEBUG
+        run: |
+          docker network create commonforgp2gp
```
ole4ryb, line 165: The command `docker network create commonforgp2gp` will fail if the network already exists. You could add a check to ensure the network only gets created if it doesn't already exist:
`docker network inspect commonforgp2gp >/dev/null 2>&1 || docker network create commonforgp2gp`

.github/workflows/build_workflow.yml
ole4ryb: Also, we can make logging levels adjustable:
`GP2GP_LOGGING_LEVEL: ${{ secrets.GP2GP_LOGGING_LEVEL || 'DEBUG' }}
 GPC_CONSUMER_LOGGING_LEVEL: ${{ secrets.GPC_CONSUMER_LOGGING_LEVEL || 'DEBUG' }}`

.github/workflows/build_workflow.yml
martin-nhs: Discussed with Adrian, happy to proceed as-is.

.github/workflows/build_workflow.yml
martin-nhs: Hey @ole4ryb - great point, though given the way that the workflow is set-up I can't see any points in which it would be _possible_ to fail, given that the network `commonforgp2gp` exists.

.github/workflows/build_workflow.yml
martin-nhs: I agree!

.github/workflows/build_workflow.yml
ole4ryb: all good

.github/workflows/build_workflow.yml
ole4ryb: just thinking more from a re-run point of view, when something fails, not sure how GitHub Actions behave it that case and whether it preserves the docker network or destroys it... probably something to explore...

.github/workflows/build_workflow.yml
martin-nhs: Hey @ole4ryb - another great point of resiliency! I have added the `-p` flag and structured the workflow a little - as for the second point I think this may be unnecessary as I think the shell is quite descriptive as-is, for example:

`cp: cannot stat './file': No such file or directory`

martin-nhs: There would be another point of `mv` being more efficient than `cp` - though the performance impact would be non-existent.

.github/workflows/build_workflow.yml
martin-nhs: So AFAIK:

* If a job fails it wouldn't use the same runner.
* As a failsafe, we have implemented:

```yml
- name: Stop Docker Dependencies
  if: always()
  run: |
    docker compose -f docker-compose.yml -f docker-compose-e2e-tests.yml down
    docker network rm commonforgp2gp
  working-directory: ./docker
```

The `if: always()` in the above example ensures that this step is always run, despite a failure - so the `commonforgp2gp` network will always be removed.

.github/workflows/build_workflow.yml
ole4ryb: Since ./dump_docker_logs.sh is a Bash script, to avoid potential issues with different shell environments, it's best to explicitly specify that it should be run using Bash:
`- name: Dump Docker Logs
        if: always()
        run: |
          chmod +x dump_docker_logs.sh
          ./dump_docker_logs.sh
        working-directory: ./scripts
        **shell: bash**`

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,333 @@
...
+        if: always()
+        with:
+          name: 'Integration Test Reports & Docker Logs'
+          path: ./artifacts/**
+          compression-level: 9
+
+      - name: Stop Docker Dependencies
+        if: always()
+        run: |
+          docker compose down --rmi=local --remove-orphans &&
```
ole4ryb, line 192: I think '&&' can be skipped ?

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,333 @@
...
+          GPC_CONSUMER_SERVER_PORT: "8090"
+          GPC_CONSUMER_SDS_URL: "http://wiremock:8080/spine-directory/"
+          GPC_CONSUMER_SDS_APIKEY: "anykey"
+          GPC_CONSUMER_LOGGING_LEVEL: "DEBUG"
+          GPC_SUPPLIER_ODS_CODE: "XYZ222"
+          GP2GP_LOGGING_LEVEL: "DEBUG"
+          MHS_MOCK_REQUEST_JOURNAL_ENABLED: "true"
+        run: |
+          docker network create commonforgp2gp
+          docker compose -f docker-compose.yml -f docker-compose-e2e-tests.yml build
```
ole4ryb, line 231: I think lines 231 and 232 can be merged into one line:
`docker compose -f docker-compose.yml -f docker-compose-e2e-tests.yml up --build --exit-code-from gp2gp-e2e-tests mongodb activemq gp2gp wiremock gpcc gp2gp-e2e-tests`

.github/workflows/build_workflow.yml
ole4ryb: sure, not a problem, I am leaving it completely up to you to assess the level of the risk.

.github/workflows/build_workflow.yml
martin-nhs: Nice spot, agreed that this should be done at a later phase! I will take note :)

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,333 @@
...
+          GP2GP_LOGGING_LEVEL: DEBUG
+          GPC_CONSUMER_LOGGING_LEVEL: DEBUG
+        run: |
+          docker network create commonforgp2gp
+          docker compose build
+          docker compose up mock-mhs-adaptor mongodb activemq wiremock gpcc gp2gp --detach
+        working-directory: ./docker
+
+      - name: Execute Integration Tests
+        run: ./gradlew integrationTest
```
ole4ryb, line 165: I'd suggest to run "./gradlew integrationTest" as a deamon (not that I am in favour of daemons by any means... :) ). But in Gradle own words the builds should run faster:
image
I run it locally and integration tests took 11 min 32 sec (without daemon) compared to 10min 32 sec (with daemon).
Without daemon:
image

With daemon:
image

martin-nhs: Good spot! Though is enabled by default since Gradle 3.0 - [source](https://docs.gradle.org/current/userguide/gradle_daemon.html#:~:text=Gradle%20enables%20the%20Daemon%20by%20default%20since%20Gradle%203.0.).

ole4ryb: yes, you are right, good finding...

ole4ryb (approved):

PR #874, opened 2024-09-16
NIAD-3173 - Fix Docker image tagging mechanism so that ECR lifecycle policies pick up orphan images

martin-nhs:
This pull request addresses an issue encountered during the testing phase of the new workflow. Specifically, the Jenkinsfile relies on the GIT_BRANCH environment variable provided by the Git plugin. To ensure compatibility with GitHub Actions, this PR implements a method to manually replicate the functionality of the GIT_BRANCH environment variable. This change facilitates the seamless integration of branch information into GitHub Actions, aligning with the workflow requirements and ensuring consistency across CI/CD platforms.

### DIFF
--- a/scripts/create_build_id.sh
+++ b/scripts/create_build_id.sh
@@ -6,17 +6,12 @@ clean_tag_element() {
 }
 
 generate_tag() {
-    local branch_name="$1"
-    local build_id="$2"
-    local git_hash="$3"
+    local clean_branch_name=$(clean_tag_element "$1")
+    local clean_build_id=$(clean_tag_element "$2")
+    local short_git_hash=$(${!3}:0:7)
 
-    local clean_branch_name
-    local clean_build_id
+    local tag="${clean_branch_name}-${clean_build_id}-${short_git_hash}"
 
-    clean_branch_name=$(clean_tag_element "$branch_name")
-    clean_build_id=$(clean_tag_element "$build_id")
-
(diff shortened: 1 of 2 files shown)


scripts/create_build_id.sh
```diff
@@ -0,0 +1,32 @@
...
+    local tag_element="$1"
+    echo "${tag_element//\//-}"
+}
+
+generate_tag() {
+    local branch_name="$1"
+    local build_id="$2"
+    local git_hash="$3"
+
+    local clean_branch_name
```
ole4ryb, line 13: I think we don't need to declated `clean_branch_name` and `lean_build_id` separately. I think these declarations can be inlined with line 16 and 17 like below:
`
local clean_branch_name=$(clean_tag_element "$branch_name")
local clean_build_id=$(clean_tag_element "$build_id")
`

scripts/create_build_id.sh
```diff
@@ -0,0 +1,32 @@
...
+    local clean_build_id
+
+    clean_branch_name=$(clean_tag_element "$branch_name")
+    clean_build_id=$(clean_tag_element "$build_id")
+
```
ole4ryb, line 18: from my perspective it would make sense to declare hash as a separate variable before using it in a tag:
`local short_git_hash="${git_hash:0:7}"`
and then use it in tag:
`echo "${clean_branch_name}-${clean_build_id}-${short_git_hash}"`

ole4ryb (approved):

PR #879, opened 2024-09-17
NIAD-3137 - Add missing test for unspecified problem

MartinWheelerMT:
## What

Add missing test to verify that an unspecified problem is produced when a related problem has an incorrect problem extension url.

## Why

The test was missing, so this existing functionality was not covered by tests

## Type of change

## Checklist:

### DIFF
--- a/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
+++ b/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
@@ -386,6 +386,25 @@ public class EncounterComponentsMapperTest {
         assertThat(removeLineEndingsAndWhiteSpace(mappedXml)).isEqualTo(removeLineEndingsAndWhiteSpace(expectedXml));
     }
 
+    @Test
+    public void When_MappingWithRelatedProblemWithIncorrectProblemExtensionUrl_Expect_UnspecifiedProblemWithOriginalText() {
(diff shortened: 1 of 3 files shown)


service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
```diff
@@ -78,6 +78,10 @@ public class EncounterComponentsMapperTest {
     private static final String INPUT_BUNDLE_TOPIC_NO_CATEGORIES = TEST_DIRECTORY + "input-bundle-17-topic-no-categories.json";
     private static final String EXPECTED_COMPONENTS_TOPIC_NO_CATEGORIES = TEST_DIRECTORY
         + "expected-components-17-topic-no-categories.xml";
+    public static final String INPUT_BUNDLE_TOPIC_RELATED_PROBLEM_INVALID_EXTENSION_URL =
+        TEST_DIRECTORY + "input-bundle-18-related-problem-invalid-problem-extension.json";
+    public static final String EXPECTED_COMPONENTS_TOPIC_RELATED_PROBLEM_INVALID_EXTENSIONS =
+        TEST_DIRECTORY + "expected-components-18-related-problem-invalid-extension.xml";
```
martin-nhs, line 84: Do these need to be static constants if they are only being used in one test? 😄

MartinWheelerMT: not really

PR #18, opened 2020-12-09
NIAD-764 Pipeline: Restore project structure, unit test reports, run tests in single stage

mattd-kainos:
* Restore /project structure
* Adds a DISABLE_TEST_CONTAINERS environment variable for integration tests

PR #881, opened 2024-09-17
NIAD-3137: Add missing test for problem extension.extension url

MartinWheelerMT:
## What

Add missing test to verify that an unspecified problem is produced when a related problem has a problem extension with an extension with an invalid url.

## Why

This functionality was present in the adaptor but was untested

## Type of change

## Checklist:

### DIFF
--- a/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
+++ b/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
@@ -387,7 +387,7 @@ public class EncounterComponentsMapperTest {
     }
 
     @Test
-    public void When_MappingWithRelatedProblemWithIncorrectProblemExtensionUrl_Expect_UnspecifiedProblemWithOriginalText() {
+    void When_MappingWithRelatedProblemWithIncorrectProblemExtensionUrl_Expect_UnspecifiedProblemWithOriginalText() {
         when(codeableConceptCdMapper.mapToCdForTopic(anyString()))
             .thenCallRealMethod();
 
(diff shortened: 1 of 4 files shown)


service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
```diff
@@ -405,6 +405,25 @@ public void When_MappingWithRelatedProblemWithIncorrectProblemExtensionUrl_Expec
             .isEqualToIgnoringWhitespace(expectedXml);
     }
 
+    @Test
+    public void When_MappingWithRelatedProblemWithIncorrectProblemExtensionExtensionUrl_Expect_UnspecifiedProblemWithOriginalText() {
```
martin-nhs, line 409:
```suggestion
    public void When_MappingWithRelatedProblemWithIncorrectProblemExtensionExtensionUrl_Expect_UnspecifiedProblemWithOriginalText() {
```

When mapping what? 😄

martin-nhs, line 409: When mapping what? 😄 (also can we make this package-private)

martin-nhs:
Looks like End-to-End tests have failed:

`#18 134.1 curl: (28) Failed to connect to archive.apache.org port 443 after 133745 ms: Could not connect to server`

May need to manually re-run!

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
MartinWheelerMT: When a related problem has an extension (with a correct problem extension url) which itself has an extension which has an incorrect URL!! (hard to fit into a test name with limited characters)
Made the added tests package private (even though it feels so wrong not having a access modifier)

MartinWheelerMT:
> Looks like End-to-End tests have failed:
> 
> `#18 134.1 curl: (28) Failed to connect to archive.apache.org port 443 after 133745 ms: Could not connect to server`
> 
> May need to manually re-run!

Yeah, it should re-run now anyway, hopefully doesn't happen again!

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterComponentsMapperTest.java
martin-nhs: I think we've discussed this on a call before, all examples provided by JUnit 5 tend to be package-private! I can see the benefit in them not being public (not totally against it), other packages shouldn't be concerned about the tests!

martin-nhs (approved):

PR #891, opened 2024-09-18
NIAD-2840: log messages optimization

ole4ryb:
## What

Logging format optimization

## Why

Logging is much more readable if it follows a uniform standard

## Type of change

## Checklist:

### DIFF
--- a/service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
+++ b/service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
@@ -52,6 +52,7 @@ public class EhrExtractStatusServiceIT {
     private static final Instant FIVE_DAYS_AGO = NOW.minus(Duration.ofDays(5));
     private static final int DEFAULT_CONTENT_LENGTH = 244;
     private static final String CONTENT_TYPE_MSWORD = "application/msword";
+    public static final String JSON_SUFFIX = ".json";
 
     @Autowired
     private EhrExtractStatusService ehrExtractStatusService;
@@ -305,7 +306,7 @@ public class EhrExtractStatusServiceIT {
                 .build())
             .gpcAccessStructured(EhrExtractStatus.GpcAccessStructured.builder()
                 .accessedAt(FIVE_DAYS_AGO)
-                .objectName(generateRandomUppercaseUUID() + ".json")
+                .objectName(generateRandomUppercaseUUID() + JSON_SUFFIX)
                 .taskId(generateRandomUppercaseUUID())
                 .build())
             .messageTimestamp(FIVE_DAYS_AGO)
@@ -393,7 +394,7 @@ public class EhrExtractStatusServiceIT {
                 .build())
             .gpcAccessStructured(EhrExtractStatus.GpcAccessStructured.builder()
                 .accessedAt(FIVE_DAYS_AGO)
-                .objectName(generateRandomUppercaseUUID() + ".json")
+                .objectName(generateRandomUppercaseUUID() + JSON_SUFFIX)
                 .taskId(generateRandomUppercaseUUID())
                 .build())
             .messageTimestamp(FIVE_DAYS_AGO)
@@ -454,7 +455,7 @@ public class EhrExtractStatusServiceIT {
                         .build())
                 .gpcAccessStructured(EhrExtractStatus.GpcAccessStructured.builder()
                         .accessedAt(FIVE_DAYS_AGO)
-                        .objectName(generateRandomUppercaseUUID() + ".json")
+                        .objectName(generateRandomUppercaseUUID() + JSON_SUFFIX)
                         .taskId(generateRandomUppercaseUUID())
                         .build())
                 .messageTimestamp(FIVE_DAYS_AGO)
--- a/service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
+++ b/service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
@@ -306,7 +306,7 @@ public class EhrExtractStatusServiceIT {
                 .build())
             .gpcAccessStructured(EhrExtractStatus.GpcAccessStructured.builder()
                 .accessedAt(FIVE_DAYS_AGO)
-                .objectName(generateRandomUppercaseUUID() + JSON_SUFFIX)
+                .objectName(generateRandomUppercaseUUIDWithJsonSuffix())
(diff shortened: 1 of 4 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -401,12 +401,14 @@ public void updateEhrExtractStatusAck(String conversationId, EhrReceivedAcknowle
         if (hasAcknowledgementExceededEightDays(conversationId, ack.getReceived())
             && hasEhrStatusReceivedAckWithUnexpectedConditionErrors(conversationId)) {
 
-            logger().warn("Received an ACK message with a conversation_id: {}, but it will be ignored", conversationId);
+            logger().warn("Received an ACK message with conversation_id: {}, "
+                          + "but it is being ignored because the EhrExtract has already been marked as a failure and 8 days have passed.",
+                          conversationId);
```
adrianclay, line 406:
```suggestion
            logger().warn("Received an ACK message with conversation_id: {}, "
                          + "but it is being ignored because the EhrExtract has already been marked as failed from not receiving an acknowledgement from the requester in time.",
                          conversationId);
```

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
```diff
@@ -305,7 +306,7 @@ private void addInProgressTransfer(String conversationId, List<EhrExtractStatus.
                 .build())
             .gpcAccessStructured(EhrExtractStatus.GpcAccessStructured.builder()
                 .accessedAt(FIVE_DAYS_AGO)
-                .objectName(generateRandomUppercaseUUID() + ".json")
+                .objectName(generateRandomUppercaseUUID() + JSON_SUFFIX)
```
martin-nhs, line 309: I think maybe having a separate method which simply returns `generateRandomUppercaseUUID() + JSON_SUFFIX` - this way you won't need to repeat `JSON_SUFFIX`.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -401,12 +401,15 @@ public void updateEhrExtractStatusAck(String conversationId, EhrReceivedAcknowle
...
+            logger().warn("Received an ACK message with conversation_id: {}, "
+                          + "but it is being ignored because the EhrExtract has already been marked as failed "
+                          + "from not receiving an acknowledgement from the requester in time.",
+                          conversationId);
             return;
         }
 
         if (hasFinalAckBeenReceived(conversationId)) {
-            logger().warn("Received an ACK message with a conversation_id=" + conversationId + " that is a duplicate");
+            logger().warn("Received an ACK message with a conversation_id: {} that is a duplicate", conversationId);
```
martin-nhs, line 412: May be out of scope for this ticket, although I am still a little confused why we need:

```java
protected Logger logger() {
    return LOGGER;
}
```
Surely adding `@Slf4j` at class-level, we can then remove this method completely - allowing up to list call `log.warn()`?

ole4ryb: I just find it as an easy way of testing logger but at the same time encapsulating it which is also a valid technique as far as I know.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
ole4ryb: ok, yes, I will crate a new method

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
martin-nhs: I understand where you're coming from, although Lombok's documentation states:

Example:
```java
 @Slf4j
 public class LogExample { }
```

will generate:
```java
public class LogExample {
   private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(LogExample.class);
}
```

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceTest.java
```diff
@@ -112,7 +112,10 @@ void shouldLogWarningWhenLateAcknowledgementReceivedAfter8DaysAndEhrReceivedAckH
         ehrExtractStatusServiceSpy.updateEhrExtractStatusAck(conversationId, ack);
 
         verify(logger, times(1))
-            .warn("Received an ACK message with a conversation_id: {}, but it will be ignored", conversationId);
+            .warn("Received an ACK message with conversation_id: {}, "
+                  + "but it is being ignored because the EhrExtract has already been marked as failed "
+                  + "from not receiving an acknowledgement from the requester in time.",
```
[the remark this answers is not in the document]
ole4ryb: I will leave it as it is as we want to be very descriptive in this case.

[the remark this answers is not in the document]
martin-nhs: I believe we could get this smaller but not lose any of the logs meaning! 😄

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -120,7 +120,7 @@ public class EhrExtractStatusService {
     private static final String ACKS_SET = ACK_HISTORY + DOT + ACKS;
     public static final int EHR_EXTRACT_SENT_DAYS_LIMIT = 8;
     private static final String UNEXPECTED_CONDITION_ERROR_CODE = "99";
-    public static final String UNEXPECTED_CONDITION_ERROR_MESSAGE = "The acknowledgement has been received after 8 days";
+    public static final String UNEXPECTED_CONDITION_ERROR_MESSAGE = "No acknowledgement has been received within 8 days";
```
adrianclay, line 123: Your functionality as it currently stands is tightly coupled to the wording of this string. You store a copy of the string in the MongoDB, and then use it to check the record matches again.

The side effect of which is that if we ever want to change this string, then we have to be aware of any existing records in the DB, and updating them to match. Otherwise the check for `hasEhrStatusReceivedAckWithUnexpectedConditionErrors` will return `false`.

adrianclay (approved):

PR #893, opened 2024-09-19
Niad-2840: Ehr Status Timeout scheduler migration

ole4ryb:
## What

The migration of Ehr Status Timeout scheduler into a new class

## Why

The scheduler should live in its own class as it has some different goals to achieve that are slightly different to wider responsibilities of EhrExtractStatusService class.

## Type of change

## Checklist:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -162,50 +145,7 @@ boolean hasEhrStatusReceivedAckWithUnexpectedConditionErrors(String conversation
...
-        if (ehrExtractStatus == null) {
-            throw new EhrExtractException(format(
-                "Couldn't update EHR received acknowledgement with error information because EHR status doesn't exist, conversation_id: %s",
-                conversationId));
-        }
-
-        logger().info("EHR status (EHR received acknowledgement) record successfully "
-                      + "updated in the database with error information conversation_id: {}", conversationId);
-
-    }
```
adrianclay, line 206: So I think the boundary we want is between code whos reason to change is due to MongoDB, and code whos reason to change is due to the business requirments of timing out transfers that haven't been acknowledged within a set amount of time.

_I think_ the code in this class should have the responsibility for MongoDB. _I also think_ that the code you've removed here is also responsible for marshaling acknowledgement data into MongoDB. If you also agree, then I'd keep them together.

ole4ryb: ok, yes, it would make sense. I will move it back to EhrExtractStatusService class

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerTest.java
```diff
@@ -0,0 +1,407 @@
...
+@ExtendWith(MockitoExtension.class)
+class EhrExtractTimeoutSchedulerTest {
+
+    public static final int NINE_DAYS = 9;
+    private static final Instant NOW = Instant.now();
+    private static final Instant FIVE_DAYS_AGO = NOW.minus(Duration.ofDays(5));
+    public static final String ACK_TYPE = "AA";
+    public static final int TWENTY_DAYS = 20;
+    private static final String UNEXPECTED_CONDITION_ERROR_CODE = "99";
+    private static final String UNEXPECTED_CONDITION_ERROR_MESSAGE = format("No acknowledgement has been received within %s days", 8);
```
martin-nhs, line 52: I think on occasions like these it helps with readability not to statically import the `format` method from `String` - but rather reference this like:

```suggestion
    private static final String UNEXPECTED_CONDITION_ERROR_MESSAGE = String.format("No acknowledgement has been received within %s days", 8);
```

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerTest.java
```diff
@@ -0,0 +1,407 @@
...
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class EhrExtractTimeoutSchedulerTest {
+
+    public static final int NINE_DAYS = 9;
+    private static final Instant NOW = Instant.now();
+    private static final Instant FIVE_DAYS_AGO = NOW.minus(Duration.ofDays(5));
+    public static final String ACK_TYPE = "AA";
+    public static final int TWENTY_DAYS = 20;
```
martin-nhs, line 50: If `TWENTY_DAYS` is only being used within `addInProgressTransfers` - can't we simplify this by having it as a local variable?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerIT.java
```diff
@@ -0,0 +1,339 @@
...
+                                     .build())
+            .messageTimestamp(FIVE_DAYS_AGO)
+            .updatedAt(FIVE_DAYS_AGO)
+            .build();
+
+        ehrExtractStatusRepository.save(extractStatus);
+    }
+
+    private @NotNull String generateRandomUppercaseUUIDWithJsonSuffix() {
+        return generateRandomUppercaseUUID() + JSON_SUFFIX;
```
martin-nhs, line 304: We don't need `JSON_SUFFIX` to be a data member!

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerIT.java
```diff
@@ -0,0 +1,339 @@
...
+import static uk.nhs.adaptors.gp2gp.ehr.EhrStatusConstants.TO_ASID;
+import static uk.nhs.adaptors.gp2gp.ehr.EhrStatusConstants.TO_ODS_CODE;
+
+@ExtendWith({MongoDBExtension.class})
+@DirtiesContext
+@SpringBootTest
+public class EhrExtractTimeoutSchedulerIT {
+
+    private static final Instant NOW = Instant.now();
+    private static final Instant FIVE_DAYS_AGO = NOW.minus(Duration.ofDays(5));
```
martin-nhs, line 42: I think this can be simplified to:

```suggestion
    private static final Instant FIVE_DAYS_AGO = Instant.now().minus(Duration.ofDays(5));
```

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerIT.java
```diff
@@ -0,0 +1,339 @@
...
+    @Autowired
+    private EhrExtractStatusRepository ehrExtractStatusRepository;
+
+    @BeforeEach
+    public void emptyDatabase() {
+        ehrExtractStatusRepository.deleteAll();
+    }
+
+    @Test
+    public void When_FindInProgressTransfers_With_OneInProgress_Expect_Returned() {
```
martin-nhs, line 57: I would try to be more explicit with the test name:

`When_FindInProgressTransfers_With_OneTransferInProgress_Expect_OneEhrExtractStatusReturned`

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerTest.java
ole4ryb: ok, I will add it

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutSchedulerTest.java
ole4ryb: I see the benefit of keeping it as static variables declared at the beginning of the class as there it is much more visible and it is easier to stop bugs by correcting them once for all places where it is used.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
ole4ryb: yes

MartinWheelerMT (approved):

PR #896, opened 2024-09-20
NIAD-2840: introducing a variable that can be changed via config file

ole4ryb:
## What

ehrExtractSentDaysLimit variable can be changed via application.yml file.

## Why

Customers can adjust the timeout duration by setting a value for EHR_EXTRACT_SENT_DAYS_LIMIT, which will then update the ehrExtractSentDaysLimit accordingly.

## Type of change

## Checklist:

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
@@ -148,7 +148,7 @@ public class EhrExtractStatusService {
         return ehrReceivedAckErrorDetailsThatContainsSearchedErrCodeAndMsg.isPresent();
     }
 
-    public boolean hasLastUpdateExceededEightDays(EhrExtractStatus ehrExtractStatus, Instant time) {
+    public boolean hasLastUpdateExceededAckTimeoutLimit(EhrExtractStatus ehrExtractStatus, Instant time) {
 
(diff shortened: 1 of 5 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
```diff
@@ -153,7 +156,7 @@ public boolean hasLastUpdateExceededEightDays(EhrExtractStatus ehrExtractStatus,
 
         long daysSinceLastUpdate = Duration.between(ehrExtractStatus.getEhrExtractCorePending().getSentAt(), time).toDays();
 
-        return daysSinceLastUpdate > EHR_EXTRACT_SENT_DAYS_LIMIT;
+        return daysSinceLastUpdate > ehrExtractSentDaysLimit;
```
stevenmccullaghmadetech, line 159: This is part of a method called "hasLastUpdateExceededEightDays"... it would seem to make sense for that method name to change too, if we're making the value variable.

stevenmccullaghmadetech (changes requested):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusService.java
ole4ryb: oh, yes, good catch @stevenmccullaghmadetech

stevenmccullaghmadetech (approved):

PR #897, opened 2024-09-20
Niad 2840: Ehr Extract status ACK err condition

ole4ryb:
## What

Ehr Status Received Ack is checked whether it includes errors

## Why

The way how we assess  whether Ehr Status Received Ack includes errors is relaxed and signals about any any errors.

## Type of change

## Checklist:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutScheduler.java
```diff
@@ -22,9 +22,9 @@
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class EhrExtractTimeoutScheduler {
 
-    private static final String UNEXPECTED_CONDITION_ERROR_CODE = "99";
-    private static final String UNEXPECTED_CONDITION_ERROR_MESSAGE
-                                            = String.format("No acknowledgement has been received within %s days", 8);
+    private static final String ERROR_CODE = "99";
+    private static final String ERROR_MESSAGE = "No acknowledgement has been received within ACK timeout limit";
```
martin-nhs, line 26: Just as a side note again, I think having some documentation around what these `ERROR_CODE`'s are (assuming that there is none) could be beneficial. Furthermore, perhaps extracting these into some enumeration?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
```diff
@@ -54,14 +60,44 @@ public class EhrExtractStatusServiceIT {
...
     @BeforeEach
     public void emptyDatabase() {
         ehrExtractStatusRepository.deleteAll();
     }
 
+    @Test
+    void updateEhrExtractStatusAckTest() {
+        var inProgressConversationId = generateRandomUppercaseUUID();
+
+        EhrExtractStatusService ehrExtractStatusServiceSpy = spy(ehrExtractStatusService);
```
martin-nhs, line 81: In terms of consistency - if using `var` let's keep this:

```suggestion
        var ehrExtractStatusServiceSpy = spy(ehrExtractStatusService);
```

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
```diff
@@ -54,14 +60,44 @@ public class EhrExtractStatusServiceIT {
...
+        ehrReceivedAcknowledgement.setReceived(NOW);
+        ehrReceivedAcknowledgement.setConversationClosed(NOW);
+
+        ehrExtractStatusServiceSpy.updateEhrExtractStatusAck(inProgressConversationId, ehrReceivedAcknowledgement);
+
+        verify(logger, times(1))
+            .warn("Received an ACK message with conversation_id: {}, "
+                  + "but it is being ignored because the EhrExtract has already been marked as failed "
+                  + "from not receiving an acknowledgement from the requester in time.",
+                  inProgressConversationId);
```
martin-nhs, line 98: Asserting against log messages in software development can lead to several issues that can affect the reliability, maintainability, and overall effectiveness of the testing process. Here are some reasons why it is generally advisable to avoid assertions based on log messages:

### Fragility of Log Messages
* Changing Formats: Log message formats can change over time due to modifications in the logging framework, changes in the application code, or updates in logging standards. Assertions against log messages can become brittle and lead to frequent test failures.
* Localization: If an application supports multiple languages, log messages may vary based on the locale. This can result in assertions failing when they shouldn't, as the expected message might not match the actual message due to translation.
### Performance Concerns
* Overhead: Extracting and asserting log messages can introduce unnecessary overhead, especially in performance-sensitive applications. This might slow down test execution and consume additional resources.
* Logging Levels: During testing, certain log levels might be disabled (e.g., debug or info), which can lead to incomplete logs and subsequent assertion failures.
### Indirect Testing of Functionality
* Testing Side Effects: Relying on log messages may lead to a focus on verifying side effects rather than the primary functionality of the code. This can result in a situation where tests pass because log messages are correct, but the actual business logic is faulty.
* Misleading Success: If a test passes because the log message is correct but the application fails to perform its intended function, it can create a false sense of security regarding the system’s correctness.
### Clarity and Intent
* Test Readability: Tests that assert log messages can be harder to read and understand, especially for new developers or team members. Clear assertions on the expected state or output of the system are generally more straightforward and intuitive.
* Focus on Requirements: Tests should ideally assert the expected behaviour or outcomes of a system rather than internal implementation details such as log messages.
### Log Messages as Implementation Details
* Encapsulation: Log messages are often implementation details that shouldn't be part of the external behaviour contract of an application. By asserting on them, tests become tightly coupled to the implementation, which is contrary to best practices in software design.
Conclusion
Instead of asserting against log messages, it is often better to verify the outcomes or state changes that occur as a result of the code execution. This approach leads to more robust tests that are less likely to break with changes in the codebase while providing greater confidence in the application's functionality.

### Conclusion
Instead of asserting against log messages, it is often better to verify the outcomes or state changes that occur as a result of the code execution. This approach leads to more robust tests that are less likely to break with changes in the codebase while providing greater confidence in the application's functionality.

### References
* Sussman, G. J., & Steele, G. L. (1980). "Scheme: An Interpreter for Extended Lambda Calculus." Proceedings of the ACM SIGPLAN 1980 Symposium on Language Issues in Programming Environments.
* B. M. H. (2016). "Effective Unit Testing: A guide for Java developers." Springer.
* Martin, R. C. (2008). "Clean Code: A Handbook of Agile Software Craftsmanship." Prentice Hall.
* Crispin, L., & Gregory, J. (2009). "Agile Testing: A Practical Guide for Testers and Agile Teams." Addison-Wesley.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
```diff
@@ -259,13 +325,7 @@ private EhrExtractStatus addCompleteTransferWithDocument() {
...
         EhrExtractStatus extractStatus = EhrExtractStatus.builder()
                 .ackHistory(EhrExtractStatus.AckHistory.builder()
-                        .acks(List.of(
-                                EhrExtractStatus.EhrReceivedAcknowledgement.builder()
-                                        .rootId(generateRandomUppercaseUUID())
-                                        .received(FIVE_DAYS_AGO)
-                                        .conversationClosed(FIVE_DAYS_AGO)
-                                        .messageRef(ehrMessageRef)
-                                        .build()))
+                        .acks(List.of(getEhrReceivedAcknowledgement(ehrMessageRef)))
```
martin-nhs, line 328: Using `Collections.singletonList()` could be beneficial here!

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutScheduler.java
ole4ryb: good suggestion, I will introduce the Enum and think about the docs.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
ole4ryb: in general and broader terms it makes sense, but in my particular case the logging with this message is the requested business logic, hence I am testing specifically for this scenario to make sure that it is triggered.

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/ErrorConstants.java
```diff
@@ -0,0 +1,23 @@
...
+        this.message = message;
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public String getMessage() {
+        return message;
+    }
```
martin-nhs, line 21: If we're using Lombok within the project, we _could_ use `@Getter` at class level - but happy for it to stay as-is, your call! 😄

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutScheduler.java
```diff
@@ -22,9 +23,7 @@
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class EhrExtractTimeoutScheduler {
 
-    private static final String ERROR_CODE = "99";
-    private static final String ERROR_MESSAGE = "No acknowledgement has been received within ACK timeout limit";
-
+    private static final ErrorConstants ACK_TIMEOUT_ERROR = ErrorConstants.ACK_TIMEOUT;
```
martin-nhs, line 26: Can't we just statically import `ACK_TIMEOUT`? 😄

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/ErrorConstants.java
```diff
@@ -0,0 +1,23 @@
+package uk.nhs.adaptors.gp2gp.ehr.utils;
+
+public enum ErrorConstants {
```
martin-nhs, line 3: I would perhaps rename this to `ErrorCode`, as Enums are constants - having `Constants` adds no value.

Here is a [reference](https://stackoverflow.com/a/3069863).

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/scheduling/EhrExtractTimeoutScheduler.java
```diff
@@ -42,8 +41,8 @@ public void processEhrExtractAckTimeouts() {
             try {
                 logger().info("Scheduler has started processing EhrExtract list with Ack timeouts");
                 ehrExtractStatusService.updateEhrExtractStatusWithEhrReceivedAckError(ehrExtractStatus.getConversationId(),
-                                                                                      ERROR_CODE,
-                                                                                      ERROR_MESSAGE);
+                                                                                      ACK_TIMEOUT_ERROR.getCode(),
+                                                                                      ACK_TIMEOUT_ERROR.getMessage());
```
martin-nhs, line 45: If you take up my suggestion regarding naming the enumeration to `ErrorCode` I would say line 26 is not required anymore and suggest accessing these like:

`ErrorCode.ACK_TIMEOUT.getCode()` etc.

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
```diff
@@ -44,24 +50,78 @@ public class EhrExtractStatusServiceIT {
...
+    @MockBean
+    private Logger logger;
+
     @BeforeEach
     public void emptyDatabase() {
         ehrExtractStatusRepository.deleteAll();
     }
 
+    @Test
+    void updateEhrExtractStatusAckTest() {
```
martin-nhs, line 80: I think this name can be improved:

* What is the tests objective?

Can we do something like `When_With_Except` here?

service/src/intTest/java/uk/nhs/adaptors/gp2gp/ehr/EhrExtractStatusServiceIT.java
```diff
@@ -44,24 +50,78 @@ public class EhrExtractStatusServiceIT {
...
+                  + "but it is being ignored because the EhrExtract has already been marked as failed "
+                  + "from not receiving an acknowledgement from the requester in time.",
+                  inProgressConversationId);
+    }
+
+    @Test
+    void updateEhrExtractStatusAckTest2() {
+        var inProgressConversationId = generateRandomUppercaseUUID();
+
+        EhrExtractStatusService ehrExtractStatusServiceSpy = spy(ehrExtractStatusService);
```
martin-nhs, line 107:
```suggestion
        var ehrExtractStatusServiceSpy = spy(ehrExtractStatusService);
```

martin-nhs (approved):

PR #924, opened 2024-10-07
Removing a variable that is not used atm

ole4ryb:
## What

Removing EHR_EXTRACT_SENT_DAYS_LIMIT variable that is not used at the moment

## Why

At the moment EHR_EXTRACT_SENT_DAYS_LIMIT that is setup in application.yml is not used hence removing it for now

## Type of change

## Checklist:

### DIFF
--- a/service/src/intTest/resources/application.yml
+++ b/service/src/intTest/resources/application.yml
@@ -14,7 +14,7 @@ management:
         include: info, health, metrics
 
 gp2gp:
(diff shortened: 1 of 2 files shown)


adrianclay (approved):
Nice! If we ever _do_ want users to configure the number we can easily add the environment variable back, plus because it's a property we could give instructions to specify it manually in an emergency.

PR #925, opened 2024-10-07
NIAD-3151: Adding more descriptive logging and code refactoring: aggregating duplicated strings into one place (string Constant Aggregator class)

ole4ryb:
Adding more descriptive logging and code refactoring: aggregating duplicated strings into one place

## What

Adding slightly more descriptive logging into DocumentReferenceToNarrativeStatementMapper class + implementing String Constant Aggregator Class

## Why

There are a lot of strings that are duplicated across many classes. They are usually defined as constants at the beginning of a class and make the class very bigger than necessary. Also, adding such string constants clutters the class and makes it less readable. By introducing the Constant Interface pattern  we aggregate such constants into one place so that the strings can be re-used between different classes. 

## Type of change

## Checklist:


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/utils/EhrConstants.java
```diff
@@ -0,0 +1,15 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class EhrConstants {
+
+    public static final String COMPLETE_CODE = "COMPLETE";
+    public static final String CONTENT_TYPE_PLACEHOLDER = "CONTENT_TYPE_PLACEHOLDER_ID=";
+    public static final String ERROR_MESSAGE_PLACEHOLDER = "ERROR_MESSAGE_PLACEHOLDER_ID=";
+    public static final String FILENAME_PLACEHOLDER = "FILENAME_PLACEHOLDER_ID=";
+    public static final String ORGANIZATION = "Organization";
+    public static final String PRACTITIONER = "Practitioner";
+}
```
adrianclay, line 15: Can you describe the relationship between thee concepts.

This seems to be a jumble of different strings.

Some of which are GP Connect resource types.
Some of which are strings we've created as part of an internal API.
Some of which are FHIR constants.

This design pattern seems to be the code equivalent of a kitchen drawer which is full of random junk which you don't know where else to put it. I hate this with a passion.

ole4ryb: The strings in this class are sourced from the ehr package, which is why I've named it EhrConstants (open to suggestions if there's a better name). My goal with this approach is to reduce the duplication of strings that are currently scattered across various classes, specifically those related to Ehr (hence located in ehr package). By consolidating these strings in one place, we create a single 'source of truth,' making refactoring easier and reducing the risk of bugs introduced by inconsistent or duplicated string values across the ehr package. By restricting the string constants class to the package level, we are narrowing its scope and improving cohesion. This addresses the previous concern raised about lack of cohesion during one of our earlier discussions. You can find more information about this approach here:
- https://medium.com/@barbieri.santiago/java-constants-best-practices-648fc562bb08
- or section "4.2. Constants Class" in this article: https://www.baeldung.com/java-constants-good-practices

PR #964, opened 2024-10-23
Improve exception messages when mapping MedicationRequest `AvailabilityTime` and `EffectiveTime` fields

MartinWheelerMT:
## What

* Improve exception message detailing the reason for the failure and the ID of the resource causing the exception. 
* Add unit tests for the `prepareAvailabilityTimeForMedicationRequest` and `prepareEffectiveTimeForMedicationRequest` methods as none previously existed.

## Why

During testing it was difficult to ascertain which resource had caused the exception, or the reasoning behind the failure.  This should make it easier to find and identify the cause of the issue.

## Type of change

## Checklist:

- [x ] I have performed a self-review of my code
- [-] My changes generate no new warnings
- [-] I have updated the [Changelog](CHANGELOG.md) with details of my change in the UNRELEASED section if this change will affect end users

### DIFF
--- a/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/utils/StatementTimeMappingUtilsTest.java
+++ b/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/utils/StatementTimeMappingUtilsTest.java
@@ -1,0 +1,120 @@
+package uk.nhs.adaptors.gp2gp.ehr.utils;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import org.assertj.core.api.ThrowableAssert;
+import org.hl7.fhir.dstu3.model.DateTimeType;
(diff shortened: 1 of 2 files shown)


service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/utils/StatementTimeMappingUtilsTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.assertj.core.api.ThrowableAssert;
+import org.hl7.fhir.dstu3.model.DateTimeType;
+import org.hl7.fhir.dstu3.model.MedicationRequest;
+
+import org.hl7.fhir.dstu3.model.Period;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import uk.nhs.adaptors.gp2gp.ehr.exception.EhrMapperException;
+
+public class StatementTimeMappingUtilsTest {
```
martin-nhs, line 15: This class probably doesn't need to be `public`!

martin-nhs (approved):

PR #980, opened 2024-10-31
NIAD-3228: removing Jackson limit of maximum size of data that it can process

ole4ryb:
## What

Removing Jackson limit of maximum size of data that it can process

## Why

ObjectMapper underneath uses Jackson which has a limit of 20 Mb of data that it can process. The data that GP2GP Adaptor is working with can be much bigger in size than the limit hence we need to remove this limit and allow the adaptor to handle much bigger documents.

## Type of change

## Checklist:


service/src/test/java/uk/nhs/adaptors/gp2gp/common/configuration/ObjectMapperConfigTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+        ObjectMapper objectMapper = context.getBean(ObjectMapper.class);
+
+        StreamReadConstraints constraints = objectMapper.getFactory().streamReadConstraints();
+
+        assertEquals(Integer.MAX_VALUE, constraints.getMaxStringLength(),
+                     "Expected maxStringLength to be set to Integer.MAX_VALUE");
+
+        context.close();
+    }
```
adrianclay, line 25: This test is wayyyy too low level in my view.

Somebody coming to this test should understand the context and the reasoning behind why this is valuable to the adaptor.

At the moment, they are told that `MaxStringLength` needs to be MAX_LENGTH. But no context as to why?!

Also, the test provides no _real_ proof that the original defect has been fixed, or that it won't regress. The test is coupled to an implementation detail as opposed to the behaviour we want our adaptor to have.

ole4ryb: this test was to address PiTest complain about the lack of tests for ObjectMapperConfig class objectMapper method. 
objectMapper method is itself very technical as it is a config bean. I can add a better description in assertEquals to give a better context... also, I will make the test name more descriptive which should give more context.

adrianclay: I don't think you have read and fully understood what I wrote Ole. Adding better description won't fix my underlying issue.

service/src/main/java/uk/nhs/adaptors/gp2gp/mhs/model/Identifier.java
```diff
@@ -1,10 +1,14 @@
...
 
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 @Builder
 @Data
+@NoArgsConstructor
+@AllArgsConstructor
```
adrianclay, line 11: What are all these about? You make no mention of why this is important in the PR. Without explaining why you've made a change it makes it very hard for other people on the team to understand what you're thinking.

ole4ryb: When I defined objectMapper as bean for the whole application it asked for default constructor in Identifier.java as without it it throws `com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `uk.nhs.adaptors.gp2gp.mhs.model.Identifier`` exception hence it added this default constructor.

adrianclay (approved):
Some suspicious line removals from random files. Not sure what's going on there!

PR #981, opened 2024-11-05
NIAD-3227: When mapping `Observation` valueQuantities, any input from the GP Connect Bundle should be escaped

MartinWheelerMT:
## What

Add functionality to XML escape the `code` and `unit` when generating the XML string. `value` does not need to be escaped as it is a `BigDecimal` type. Add unit tests to test above functionality.

## Why

To avoid the possibility of XML Injection, and to ensure that any generated XML is formatted correctly, all input from the GP Connect Bundle should be XML escaped when mapping `Observation` valueQuantities.

For example, if the input bundle contains the following valueQuantity:

```json
"valueQuantity": {
    "value": 186,
    "unit": "<None>",
    "system": "http://unitsofmeasure.org"
}
```
Then the resultant `<value>` element should contain the XML Escaped value for `unit`:

```xml
<value xsi:type="PQ" value="186" unit="1">
    <translation value="186">
        <originalText>&lt;None&gt;</originalText>
    </translation>
</value>
```

## Type of change

## Checklist:

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationValueQuantityMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationValueQuantityMapper.java
@@ -1,6 +1,7 @@
 package uk.nhs.adaptors.gp2gp.ehr.mapper;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.StringEscapeUtils;
 import org.hl7.fhir.dstu3.model.BooleanType;
 import org.hl7.fhir.dstu3.model.Extension;
 import org.hl7.fhir.dstu3.model.Quantity;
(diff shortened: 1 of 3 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationValueQuantityMapper.java
```diff
@@ -68,9 +72,9 @@ private static String getPhysicalQuantityXml(Quantity valueQuantity) {
                 .formatted(
                     valueQuantity.getValue(),
                     valueQuantity.getValue(),
-                    valueQuantity.getCode(),
+                    escapedCode,
                     getSystemWithoutPrefix(valueQuantity.getSystem()),
```
adrianclay, line 76: Should we be worried about the system here at all? It's a string.

MartinWheelerMT: [This method](https://github.com/NHSDigital/integration-adaptor-gp2gp/blob/0d00ec63f4b9accd1af7b13e1da5ed5dccef7683/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/ObservationValueQuantityMapper.java#L213) validates the system before it is used.  So if it isn't specifically `http://unitsofmeasure.org`, an OID, or a URN, then it isn't used.  As these don't contain XML escape characters we don't need to escape it.

adrianclay (approved):

PR #982, opened 2024-11-06
NIAD-3222: GP2GP adaptor generates an empty `CompoundStatement/code` which fails the transfer.

MartinWheelerMT:
## What

* Replace String.Empty when preparing the code element mapping ObservationMapper so that an EhrMapperException is thrown containing the resource identifier. 
* Add test to ensure this exception is found and contains the resource ID.
* Update NOPAT test and associated resource file which previously did not contain a code element 
* Update CHANGELOG.md

## Why

Within the `diagnosticreport.ObservationMapper::prepareCodeElement` there is the possibility of generating an EMPTY code element:

```java
private String prepareCodeElement(Observation observation) {
    if (observation.hasCode()) {
        return codeableConceptCdMapper.mapCodeableConceptToCd(observation.getCode());
    }

    return StringUtils.EMPTY;
}
```

This results in a `CompoundStatement` without a code element, which is invalid, and will cause the requesting system to reject the EHRExtract.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
@@ -488,7 +488,7 @@ public class ObservationMapper {
             return codeableConceptCdMapper.mapCodeableConceptToCd(observation.getCode());
         }
 
-        return StringUtils.EMPTY;
+        throw new EhrMapperException("%s must contain a code element.".formatted(observation.getId()));
     }
 
     private CompoundStatementClassCode prepareClassCode(List<MultiStatementObservationHolder> derivedObservations) {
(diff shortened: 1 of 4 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
```diff
@@ -488,7 +488,7 @@ private String prepareCodeElement(Observation observation) {
             return codeableConceptCdMapper.mapCodeableConceptToCd(observation.getCode());
         }
 
-        return StringUtils.EMPTY;
+        throw new EhrMapperException("%s must contain a code element.".formatted(observation.getId()));
```
adrianclay, line 491: Does `getId()` return something like `Observation/abc` ? If so then that looks good to me.

service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapperTest.java
```diff
@@ -218,10 +220,21 @@ void When_MappingTestGroupHeader_With_NopatMetaSecurity_Expect_ConfidentialityCo
 
         assertAll(
             () -> assertThatXml(actualXml).containsXPath(COMPOUND_STATEMENT_CONFIDENTIALITY_CODE_XPATH),
-            () -> assertThat(metaWithNopat).hasSize(1)
```
adrianclay, line 221: I'm a bit confused as to why this has changed from 1 to 2.

adrianclay (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/diagnosticreport/ObservationMapper.java
adrianclay: Checked, and it does.

PR #1000, opened 2024-11-20
Code Quality: Replace deprecated constructor call and fix unchecked assignment warning

MartinWheelerMT:
## What

* Code Quality fix to replace deprecated constructors on `URL(java.lang.String) with the recommended `URI().toURL()` method instead 
* Fix warning with Unchecked assignment: 'java.util.HashMap' to 'java.util.Map<java.lang.String,java.lang.Object>'"
 
## Why

[Deprecate URL public constructors](https://bugs.openjdk.org/browse/JDK-8294241)
Warning: Unchecked assignment: 'java.util.HashMap' to 'java.util.Map<java.lang.String,java.lang.Object>'

## Type of change

## Checklist:

### DIFF
--- a/wiremock/src/main/java/PatientDemographicsServiceClient.java
+++ b/wiremock/src/main/java/PatientDemographicsServiceClient.java
@@ -1,5 +1,6 @@
 import com.auth0.jwt.JWT;
 import com.auth0.jwt.algorithms.Algorithm;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Charsets;
 import org.apache.hc.client5.http.utils.Base64;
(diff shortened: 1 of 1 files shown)


wiremock/src/main/java/PatientDemographicsServiceClient.java
```diff
@@ -37,7 +38,7 @@ public Map<String, Object> patient(String nhsNumber) throws Exception {
 
         String patientData = getPatientData(nhsNumber, accessToken);
 
-        return new ObjectMapper().readValue(patientData, HashMap.class);
+        return new ObjectMapper().readValue(patientData, new TypeReference<>() {});
```
MartinWheelerMT, line 41: The part which is new TypeReference<>() {} is actually new TypeReference<Map<String, Object>>() {}  but you can omit the Map<String, Object>> if it can be inferred.

This is a way to provide full type information to Jackson when deserializing, avoiding unchecked warnings. It's a generic safe way to tell the object mapper what it is being deserialized into

adrianclay (approved):

PR #1024, opened 2024-12-05
NIAD-3234: endpoint docs update

ole4ryb:
## What

Endpoints documentation update

## Why

The documentation must remain synchronized with the code implementation to ensure accuracy, clarity, and ease of use. This alignment minimizes confusion, reduces the risk of errors during integration, and provides developers with reliable information to effectively utilize the API.

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,6 @@ GP2GP 2.2b producer, or those wishing to decommission their existing producer, m
 ## How to query the EHR Status API and the Requests endpoint
 
 An API is provided to query the status of any transfer to an incumbent.
-Any "in-progress" transfers are excluded from this list, but become available once they either succeed or fail.
 
 The API has been documented using the OpenAPI Specification (version 3.0.1). The documentation includes detailed descriptions of endpoints,
 parameters, and data models, as well as examples of requests and responses to help developers integrate and use the API effectively
--- a/README.md
+++ b/README.md
@@ -2,16 +2,16 @@
 
 National Integration Adaptor - [GP2GP Sending Adaptor](https://digital.nhs.uk/developer/api-catalogue/gp2gp/gp2gp-sending-adaptor)
 
-The existing GP2GP solution uses a legacy messaging standard and infrastructure (HL7v3 and Spine TMS). Reliance 
-on these standards going forward presents a significant barrier to successful GP2GP implementation by new suppliers, 
-and perpetuation of these standards in the long term presents a risk to the continued operation of GP2GP across all 
+The existing GP2GP solution uses a legacy messaging standard and infrastructure (HL7v3 and Spine TMS). Reliance
+on these standards going forward presents a significant barrier to successful GP2GP implementation by new suppliers,
+and perpetuation of these standards in the long term presents a risk to the continued operation of GP2GP across all
 suppliers.
 
-A hybrid solution approach has been selected as the best option for GP2GP adoption by NMEs and transition by existing 
+A hybrid solution approach has been selected as the best option for GP2GP adoption by NMEs and transition by existing
 incumbent suppliers.
 
-The "National Integration Adaptor - GP2GP" implements a GP2GP 2.2b producer using the supplier's existing GP Connect 
-Provider implementation to extract the Electronic Health Record. Suppliers that have not already implemented a 
+The "National Integration Adaptor - GP2GP" implements a GP2GP 2.2b producer using the supplier's existing GP Connect
+Provider implementation to extract the Electronic Health Record. Suppliers that have not already implemented a
 GP2GP 2.2b producer, or those wishing to decommission their existing producer, may deploy the GP2GP adaptor in its place.
 
 ## Table of contents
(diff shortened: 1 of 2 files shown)


gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+        post:
+            tags:
+                - Ehr Status requests controller
+            operationId: getEhrRequestsEncodedForm
+            parameters:
+                - name: request
+                  in: query
+                  required: true
+                  schema:
+                      $ref: '#/components/schemas/EhrStatusRequestQuery'
```
adrianclay, line 19: For my understanding, is this saying you can pass the filtering parameters in via query parameters? Also it's a bit strange it's marked as required, given that I don't think I've ever used the `/requests` endpoint in this way. 🤔

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                            request-example:
+                                summary: Example Request for EHR Status
+                                value:
+                                    fromDateTime: "2020-10-31T01:30:00.000Z"
+                                    toDateTime: "2030-10-31T01:30:00.000Z"
+                                    fromAsid: "918999198738"
+                                    toAsid: "200000000359"
+                                    fromOdsCode: "P883302"
+                                    toOdsCode: "P113302"
+                required: true
```
adrianclay, line 35: Hmm, I'm pretty sure I've also called this endpoint without specifying a request body and it's worked. 🤔

README.md
```diff
@@ -21,276 +21,14 @@ GP2GP 2.2b producer, or those wishing to decommission their existing producer, m
...
-    "fromAsid": "918999198738",
-    "toAsid": "200000000359"
-}
-```
-</details>
-
-## How to query the Requests endpoint
-
-An API is provided to query the status of all transfers to an incumbent.
 Any "in-progress" transfers are excluded from this list, but become available once they either succeed or fail.
```
adrianclay, line 27: Is there a reasonable place for this to live within the `yml` too?

README.md
```diff
@@ -21,276 +21,14 @@ GP2GP 2.2b producer, or those wishing to decommission their existing producer, m
...
 
-Requests can be made to the following endpoint, where each attribute within the JSON POST body is an optional filter criteria.
```
adrianclay, line 234: Related to my comments below about `required`, this comment reaffirms my understanding which is that the filtering is optional.

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                  required: true
+                  schema:
+                      type: string
+            responses:
+                '202':
+                    description: Successfull Ehr Extract resend request
+                    content:
+                        application/json:
+                            schema:
+                                type: string
```
adrianclay, line 91: I don't know if there is a way to explicitly say that the response is empty?

Example response headers:
```
HTTP/1.1 202 
Date: Fri, 29 Nov 2024 10:58:26 GMT
Content-Length: 0
Connection: keep-alive
```

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                '409':
+                    description: Ehr Extract resend action cannot be triggered
+                    content:
+                        application/json:
+                            schema:
+                                type: string
```
adrianclay, line 103: For both this and the one above, it would be nice if there was a way to specify that an `OperationOutcome` is returned.

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                - name: conversationId
+                  in: path
+                  required: true
+                  schema:
+                      type: string
+            responses:
+                '200':
+                    description: OK
+                    content:
+                        '*/*':
```
adrianclay, line 119: Any reason not to use `application/json` here?

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                    type: string
+                toAsid:
+                    type: string
+        ErrorDetails:
+            type: object
+            properties:
+                code:
+                    type: string
+                display:
+                    type: string
```
adrianclay, line 546: A bit sparse, but not the end of the world. Nullable: false would be good.

README.md
```diff
@@ -21,276 +21,14 @@ GP2GP 2.2b producer, or those wishing to decommission their existing producer, m
...
-            "messageRef": "E600D1D4-D5BE-4D0D-8080-F7F6188B0548"
-        }
-    ],
-    "migrationStatus": "FAILED_INCUMBENT",
-    "originalRequestDate": "2023-07-24T10:42:05.757Z",
-    "fromAsid": "918999198738",
-    "toAsid": "200000000359"
-}
-```
-#### Failed by adaptor, due to a PATIENT_NOT_FOUND error from the GP Connect provider's migrate structured endpoint
```
adrianclay, line 215: Did we lose this example?

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                error:
+                    $ref: '#/components/schemas/Error'
+                messageTimestamp:
+                    type: string
+                    format: date-time
+                ehrExtractMessageId:
+                    type: string
+                ackHistory:
+                    $ref: '#/components/schemas/AckHistory'
+            description: Details about EhrExtractStatus
```
adrianclay, line 398: Does this schema definition get used anywhere? It also references schema types that are only referenced by this.

README.md
ole4ryb: ok, I see, make sense, I will look into it and adjust it

gp2gp_adaptor_response_docs.yaml
```diff
@@ -0,0 +1,557 @@
...
+                fileName:
+                    type: string
+                    nullable: False
+                    description: 'The filename sent to winning practice in the GP2GP message'
+                originalDescription:
+                    type: string
+                    nullable: True
+                    description: 'The description of the file as given by the description element of the GP Connect DocumentReference resource.
+
+            This is inserted into the GP2GP placeholder document as the original filename and suffix. So should be of the form filename.suffix, e.g. referral.txt'
```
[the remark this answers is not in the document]
ole4ryb: done

gp2gp_adaptor_response_docs.yaml
ole4ryb: done

gp2gp_adaptor_response_docs.yaml
ole4ryb: I've adjusted it

gp2gp_adaptor_response_docs.yaml
ole4ryb: I've added it

gp2gp_adaptor_response_docs.yaml
ole4ryb: I've removed those schemas

README.md
ole4ryb: I think as object it is required but EhrStatusRequestQuery object has fields which are not mandatory and can be empty.

adrianclay (approved):

PR #1031, opened 2024-12-18
Prevent Sonar Reports and Mutation Testing running when triggered by Dependabot

MartinWheelerMT:
## What

* Add an if condition to the mutation testing GitHub action to prevent the mutation testing from being run if the GitHub actor is dependabot.
* Add a conditional job for the build workflow to not generate a sonar report when triggered by dependabot.

## Why

Currently the mutation testing workflow triggers when dependabot raises a PR, and generating the Sonar report also fails in the unit tests stage of the build action.
These fail as Dependabot does not have access to GitHub secrets.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -107,16 +107,15 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
 
       - name: Execute Unit Tests
-        if: github.actor != 'dependabot[bot]'
-        run: ./gradlew test jacocoTestReport sonar --parallel --build-cache
+        run: |
+          if [[ $GITHUB_ACTOR == 'dependabot[bot]' ]]; then
+            ./gradlew test jacocoTestReport --parallel --build-cache
+          else
+            ./gradlew test jacocoTestReport sonar --parallel --build-cache
+          fi
         working-directory: ./service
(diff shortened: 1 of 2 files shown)


.github/workflows/build_workflow.yml
```diff
@@ -107,11 +107,17 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
 
       - name: Execute Unit Tests
+        if: github.actor != 'dependabot[bot]'
```
adrianclay, line 110: I'm not sure about this. We surely want some confidence that a dependency bump doesn't break CI?

Skipping mutation testing makes sense, as dependabot doesn't change our production code for which the mutations would be generated for.

Skipping our unit tests is scary though.

adrianclay: Just seen the addition below. Is there a way to combine the two jobs? Conditionally add 'sonar' to the list of jobs depending on the actor.

MartinWheelerMT: Yeah,  I wouldn't have skipped the unit tests!

I think that the change I've made should work now with the two conditional steps moved into one and using a conditional in the run script

ole4ryb (approved):

PR #1054, opened 2025-01-13
NIAD-3208: CustomTrustStore refactoring to exclude weird dependencies

ole4ryb:
## What

## Why

## Type of change

## Checklist:

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -133,7 +133,6 @@ tasks.withType(Test) {
 }
 
 check.dependsOn integrationTest
-integrationTest.mustRunAfter test
 
 spotbugsTest.enabled = false
 spotbugsIntegrationTest.enabled = false
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -133,6 +133,7 @@ tasks.withType(Test) {
 }
 
 check.dependsOn integrationTest
+integrationTest.mustRunAfter test
 
 spotbugsTest.enabled = false
 spotbugsIntegrationTest.enabled = false
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/AppInitializer.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/AppInitializer.java
@@ -3,7 +3,6 @@ package uk.nhs.adaptors.gp2gp.common.configuration;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.Bean;
 import org.springframework.stereotype.Component;
 
 import lombok.RequiredArgsConstructor;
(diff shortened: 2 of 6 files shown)


service/build.gradle
```diff
@@ -133,6 +133,7 @@ tasks.withType(Test) {
 }
 
 check.dependsOn integrationTest
+integrationTest.mustRunAfter test
```
adrianclay, line 136: What's this change all about?

I'm not sure I'd consider this a refactoring.

service/src/test/java/uk/nhs/adaptors/gp2gp/common/configuration/CustomTrustStoreTest.java
```diff
@@ -48,6 +47,11 @@ static void setUp() {
                            software.amazon.awssdk.core.sync.RequestBody.fromFile(trustStoreFile));
     }
 
+    @BeforeEach
+    void prepare() {
+        customTrustStore = new CustomTrustStore(s3Client);
```
adrianclay, line 52: Minor, why not put it in the existing `@BeforeAll` ?

ole4ryb: yes, I can do

service/build.gradle
ole4ryb: I will remove this line

service/src/test/java/uk/nhs/adaptors/gp2gp/common/configuration/AppInitializerTest.java
```diff
@@ -0,0 +1,40 @@
...
+    @Test
+    void getNullWhenTrustStoreUrlDoesNotStartWithS3Prefix() {
+
+        storageConnectorConfiguration.setTrustStoreUrl("http://localhost");
+        appInitializer = new AppInitializer(storageConnectorConfiguration);
+
+        S3Client s3Client = appInitializer.getS3Client();
+
+        assertNull(s3Client);
+    }
```
adrianclay, line 39: Thinking about this I don't think this behaviour is actually useful. The CustomTrustStore needs an S3Client to work, so having a CustomTrustStore created but without an S3Client is buggy.

ole4ryb: But in turn the creation of s3Client is depended on the trustStoreUrl and if that's not provided then s3Client also doesn't seem to be useful for the trustStoreEven if constructed... Looks like I will have to change it and raise an exception instead of returning null

service/src/main/java/uk/nhs/adaptors/gp2gp/common/configuration/AppInitializer.java
```diff
@@ -32,13 +33,12 @@ public void afterPropertiesSet() {
...
-    @Bean
-    public S3Client getS3Client() {
+    public S3Client getS3Client() throws ConfigurationException {
         if (StringUtils.isNotBlank(storageConnectorConfiguration.getTrustStoreUrl())
             && storageConnectorConfiguration.getTrustStoreUrl().startsWith(S3_PREFIX)) {
             return S3Client.builder().build();
         }
 
-        return null;
+        throw new ConfigurationException("S3Client cannot be instantiated due to trust store URL misconfiguration");
```
adrianclay, line 42: Any chance this error message can be made more specific? This is likely going to be shown in the error logs, and as an NME I'd have no idea how to fix this.

ole4ryb: thanks, I've adjusted the log message

adrianclay (approved):

PR #1061, opened 2025-01-21
Upgrade minimum version of MongoDB required for the adaptor

adrianclay:
⚠️ Will need to think about how we deal with this, is a breaking change to upgrade the minimum version ⚠️ 

## Type of change

Internal change (non-breaking change with no effect on the functionality affecting end users)

## Checklist:


adrianclay:
Checked with Medicus and they won't be affected by this.

stevenmccullaghmadetech (approved):
I searched the codebase for any breaking changes between these MongoDB versions  -looks fine. I made a small typo change to the OPERATING.md file.

PR #1146, opened 2025-03-10
NIAD-3304: Adaptor not preserving legacy read codes on export

MartinWheelerMT:
## What

* Add tests for existing functionality when mapping `CodeableConcept` for blood pressures.
* Slight refactor on `CodeableConceptCdMapper` for readability and allow testing around existing bug when retrieving the original text value from the display extension.

## Why

This method was untested and changes are required when handling non-SNOMEDCT codes.

## Type of change

## Checklist:


MartinWheelerMT:
Pitest surviving mutant is due to a bug in the original code which cannot be tested as the code is never hit.  This bug will be addressed in a future pull request

Bhavesh-Patel (approved):

PR #1168, opened 2025-03-20
NIAD-3304: Ensure correct ordering of `originalText` when mapping `CodeableConcept` to `CD`

MartinWheelerMT:
## What

* Update CD template to ensure that `originalText` is always populated first.
* Update tests to reflect the ordering change
* Update `CodeableConceptCdMapper` to only add code translations when it is possible to map from URL to OID as anything other than an OID is not supported in the schema.
* Update CodesSystemsUtil to return empty string if a mapping to known code system cannot be found.
* Update test accordingly.
* Update `CodeableConceptCdMapper` to create a new coding element for legacy codes, to avoid modifying an element whilst iterating a list.
* Add test for an unknown code system.
* SonarCube fix to remove public modifier on test class.
* Fix error in "Medications" test file, where a legacy code contains a space, which is not valid according to the schema.

## Why

The imports were failing on EMIS due to schema validation issues, looking at the schema:

```xml
   <xsd:complexContent>
            <xsd:extension base="ANY">
                <xsd:sequence>
                    <xsd:element name="originalText" type="ED" minOccurs="0">
                        <xsd:annotation>
                            <xsd:documentation> The text or phrase used as the basis for the coding. </xsd:documentation>
                            <xsd:appinfo />
                        </xsd:annotation>
                    </xsd:element>
                    <xsd:element name="qualifier" type="CR" minOccurs="0" maxOccurs="unbounded">
                        <xsd:annotation>
                            <xsd:documentation> Specifies additional codes that increase the
                                specificity of the primary code. </xsd:documentation>
                            <xsd:appinfo />
                        </xsd:annotation>
                    </xsd:element>
                    <xsd:element name="group" minOccurs="0" maxOccurs="unbounded">
                        <xsd:complexType>
                            <xsd:sequence>
                                <xsd:element name="qualifier" type="CR" maxOccurs="unbounded" />
                            </xsd:sequence>
                        </xsd:complexType>
                    </xsd:element>
                    <xsd:element name="translation" type="CD" minOccurs="0" maxOccurs="unbounded">
                        <xsd:annotation>
                            <xsd:documentation> A set of other concept descriptors that translate
                                this concept descriptor into other code systems. </xsd:documentation>
                            <xsd:appinfo />
                        </xsd:annotation>
                    </xsd:element>
                </xsd:sequence>
...
```
It appears that the `originalText` field should be appearing first in the sequence ahead of the translations.

## Type of change

## Checklist:

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
@@ -1,5 +1,6 @@
 package uk.nhs.adaptors.gp2gp.ehr.mapper;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
@@ -316,12 +317,17 @@ public class CodeableConceptCdMapper {
             .filter(coding -> !isSnomed(coding))
             .toList();
 
+        List<Coding> nonSnomedCodings = new ArrayList<>();
+
         for (Coding coding : nonSnomedCodeCodings) {
             var hl7CodeSystem = CodeSystemsUtil.getHl7code(coding.getSystem());
-            coding.setSystem(hl7CodeSystem);
+            if (!hl7CodeSystem.isEmpty()) {
+                coding.setSystem(hl7CodeSystem);
+                nonSnomedCodings.add(coding);
+            }
         }
 
-        return nonSnomedCodeCodings;
+        return nonSnomedCodings;
     }
 
     private Optional<String> findOriginalText(CodeableConcept codeableConcept, Optional<Coding> coding) {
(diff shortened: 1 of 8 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/ehr/mapper/CodeableConceptCdMapper.java
```diff
@@ -316,12 +317,16 @@ private List<Coding> getNonSnomedCodeCodings(CodeableConcept codeableConcept) {
             .filter(coding -> !isSnomed(coding))
             .toList();
 
+        List<Coding> nonSnomedCodings = new ArrayList<>();
+
         for (Coding coding : nonSnomedCodeCodings) {
             var hl7CodeSystem = CodeSystemsUtil.getHl7code(coding.getSystem());
-            coding.setSystem(hl7CodeSystem);
+            if (!hl7CodeSystem.isEmpty()) {
+                nonSnomedCodings.add(new Coding(hl7CodeSystem, coding.getCode(), coding.getDisplay()));
```
MartinWheelerMT, line 325: This was needed as an issue was occurring where the same coding block was processed twice when it was updated whilst iterating the list

PR #1189, opened 2025-04-03
NIAD-3156: Send over Encounter.meta.security NOPAT field to incumbent

ole4ryb:
## What

Send over Encounter.meta.security NOPAT field to incumbent

## Why

Making Encounter resource compliant with Redactions changes

## Type of change

## Checklist:

### DIFF
--- a/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapperTest.java
+++ b/service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapperTest.java
@@ -10,6 +10,7 @@ import static org.mockito.Mockito.when;
 import java.time.Instant;
 import java.util.Date;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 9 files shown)


service/src/test/java/uk/nhs/adaptors/gp2gp/ehr/mapper/EncounterMapperTest.java
```diff
@@ -80,14 +91,32 @@ public void setUp() {
...
     }
 
     @AfterEach
     public void tearDown() {
         messageContext.resetMessageContext();
     }
 
+    @Test
+    public void testEncounterWithNOPATAddsConfidentialityCodeIntoEhrComposition() {
+        lenient().when(randomIdGeneratorService.createNewId()).thenReturn(TEST_ID);
```
MartinWheelerMT, line 104: not a change, just some info.  You only need the lenient() when there will be cases where mocked service will not occur, so for this test it shouldn't need to be included

MartinWheelerMT (approved):

PR #1503, opened 2026-07-06
Adding additional log coverage

ORybak5:
## What

Adding additional log coverage

## Why

Log coverage is an essential part of the application and helps to investigate different issues and cases.

## Type of change

## Checklist:

### DIFF
--- a/service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcClient.java
+++ b/service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcClient.java
@@ -23,33 +23,67 @@ public class GpcClient {
     private final GpcRequestBuilder gpcRequestBuilder;
 
     public String getStructuredRecord(GetGpcStructuredTaskDefinition structuredTaskDefinition) {
-        String gpcBaseUrlWithOds = buildGpcBaseUrl(structuredTaskDefinition);
-        var requestBody = gpcRequestBuilder.buildGetStructuredRecordRequestBody(structuredTaskDefinition);
-        var request = gpcRequestBuilder.buildGetStructuredRecordRequest(requestBody, structuredTaskDefinition, gpcBaseUrlWithOds);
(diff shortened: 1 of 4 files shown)


service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcClient.java
```diff
@@ -23,33 +23,67 @@ public class GpcClient {
...
+            var requestBody = gpcRequestBuilder.buildGetStructuredRecordRequestBody(structuredTaskDefinition);
+            var request = gpcRequestBuilder.buildGetStructuredRecordRequest(requestBody, structuredTaskDefinition, gpcBaseUrlWithOds);
 
-        if (LOGGER.isDebugEnabled()) {
-            var responseBody = performRequest(request);
-            LOGGER.debug("Get StructuredRecord response body: {}", LogSanitizer.sanitize(responseBody));
+            logRequest(STRUCTURED_LOG_TEMPLATE, structuredTaskDefinition,
+                gpcConfiguration.getUrl() + gpcConfiguration.getMigrateStructuredEndpoint());
+
+            long startTime = System.currentTimeMillis();
```
chiaramapellimt, line 36: random Q but why do we use long here?

ORybak5: Miliseconds quite be quite long hence safer to declare them as "long" rather than as "int".

chiaramapellimt (approved):

service/src/main/java/uk/nhs/adaptors/gp2gp/gpc/GpcClient.java
MartinWheelerMT: main reason is:

This method returns the difference, measured in milliseconds, between the current time and midnight, January 1, 1970 UTC(coordinated universal time). - Thats a pretty big number so the method returns long as int has a max value of 2147483647

PR #1511, opened 2026-07-21
Fixing ContactPoint NPE

ORybak5:
## What

Fixing ContactPoint NPE

## Why

The code should be stable and resilient against NullPointerExceptions. Instead of throwing an NPE, it will throw a dedicated exception that more clearly identifies the root cause of the issue.

## Type of change

## Checklist:


MartinWheelerMT (approved):

chiaramapellimt (approved):
looks good to me and martin YO

